PR #90, opened 2023-09-13
[draft] example specs

soenkehahn:


soenkehahn:
This is pretty outdated and we also stole some of this into other PRs. Closing.

PR #160, opened 2023-10-09
Sundry ideas

jkarni:
When writing docs, I started digging into things a bit more to be able to speak intelligently about them. I ended up playing around quite a bit with things, and thought I'd throw all of it in a PR so maybe we can discuss them.


jkarni:
Closing this since a lot has changed.

PR #235, opened 2023-10-13
Properly support `enter` & `check` in `mkNpmProject`

### LINKED ISSUE #229: Figure out an easy way to run `tsc` as a `Check` and within `garn enter`
#227 demonstrates a few issues that we would like to work. Maybe there's some changes to how people can achieve them, but currently it's not clear these things are doable at all. This includes being able to do:

- `garn enter default` and then `tsc` (or `npm install ; tsc`),
- Putting .addCheck("tsc")`tsc` into the `garn.ts` file and then having it run with `garn check default`.

soenkehahn:
Fixes #229.


examples/monorepo/garn.ts
```diff
@@ -18,11 +18,9 @@ export const npmFrontend = garn.javascript.mkNpmProject({
...
   src: "frontend-npm",
   nodeVersion: "18",
-  testCommand: "npm run test -- --watchAll=false",
 });
 
 export const startAll = garn.processCompose({
   backend: backend.defaultExecutable!,
-  'yarn frontend': yarnFrontend.defaultExecutable!,
-  'npm frontend': npmFrontend.defaultExecutable!,
+  frontend: npmFrontend.devShell.shell`cd frontend-npm && npm install && npm start`,
```
soenkehahn, line 25: We don't run the yarn frontend here anymore?

soenkehahn (commented):
Apart from my one comment, this LGTM for now. What do you think, @alexdavid?

examples/monorepo/garn.ts
alexdavid: I figured maybe it was best to omit it for now until our yarn API becomes more stable. Happy to add it back if you feel differently though!

alexdavid (approved):
This looks great!

examples/monorepo/garn.ts
soenkehahn: I think you can do ```npmFrontend.shell`...```.

soenkehahn: And maybe replace the yarn project with the haskell one?

PR #242, opened 2023-10-15
First pass at Concepts doc

### LINKED ISSUE #166: Write a `Concepts` page
Explain `Project`, `Check`, etc.

jkarni:
Relates to, and maybe fixes #166

### DIFF
--- a/website/src/docs/concepts.mdx
+++ b/website/src/docs/concepts.mdx
@@ -4,5 +4,119 @@ export const info = {
   index: 1
 }
 
+import { Hovernote } from "/src/components/Hovernote.tsx"
+
 # Basic concepts
 
+There are only a few concepts that are needed to understand <b>garn</b>.
+The concepts are tied to the capabilities provided by the <b>garn</b> CLI.
+
+|               | Environment  |  Package   |  Check      | Executable |  Project
+|---------------|--------------|------------|-------------|------------|---------
+| command       | enter        | build      | check       | run        | --help
+| is sandboxed  | can be       | yes        | yes         | no         | --help
+
+
+## Environment
+
+An *environment* is the context in which a program runs. That includes things
+like environment variables, current working directory, state of the filesystem,
+networking stack, etc. It's what makes running the exact same program not always
+result in the exact same thing.
+
+`garn` makes creating new environments quite easy. You can easily create an
+environment where some files (for instance, executables and packages) are
+available, and variables such as PATH are set.
+
+Importantly, environments can be combined! When you run `garn enter foo`, you are
+combining your *current environment* with the environment `foo`.
+
+Two environments are particularly worth note. One we just mentioned - your
+*current environment* - i.e., the environment in which a `garn` command is run.
+The other is the *empty* one with no networking, no files, no environment
+variables. Sandboxes tend to approximate this environment. Checks and builds
+run in a combination of this empty environment and the environment you specify
+in the check.
+
+If `myEnv` has type `Environment`, you get two methods on `myEnv` that relate
+to these aforementioned environments:
+
+- `myEnv.shell`: This is a template literal that runs the provided command in
+  the combination of `myEnv` and the current environment. This is useful for
+  development commands.
+- `myEnv.check`: This is a template literal that runs the provided command in
+  the combination of `myEnv` and the empty environment. Because it doesn't allow
+  access to the current environment, it is reproducible: if you run it twice,
+  the result will be the same<Hovernote>So much so that, as an optimization,
+  `garn` and `nix` store the result of the first run to avoid rerunning a second
+  time.</Hovernote>. And if you run it on different computers, then, with some
+  caveats, the result will be the same<Hovernote>So much so that, as an optimization,
+  if you run it a CI like <a href="https://garnix.io">garnix</a> and set it up
+  as a cache, and then run it locally, the result obtained by **garnix** will
+  be downloaded and reused, rather than run locally</Hovernote>
+
+
+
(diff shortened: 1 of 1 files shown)


website/src/docs/concepts.mdx
```diff
@@ -4,5 +4,119 @@ export const info = {
...
+like environment variables, current working directory, state of the filesystem,
+networking stack, etc. It's what makes running the exact same program not always
+result in the exact same thing.
+
+`garn` makes creating new environments quite easy. You can easily create an
+environment where some files (for instance, executables and packages) are
+available, and variables such as PATH are set.
+
+Importantly, environments can be combined! When you run `garn enter foo`, you are
+combining your *current environment* with the environment `foo`.
```
soenkehahn, line 32: Personally I don't find it very helpful to think of my current environment as an `Environment` in the `garn` sense. Because it's not really something that could ever be captured by a value of type `Environment`. And I think I would find it more helpful to have this document stick to what the `Environment` type is and allows users to do.

website/src/docs/concepts.mdx
```diff
@@ -4,5 +4,119 @@ export const info = {
...
+  the combination of `myEnv` and the current environment. This is useful for
+  development commands.
+- `myEnv.check`: This is a template literal that runs the provided command in
+  the combination of `myEnv` and the empty environment. Because it doesn't allow
+  access to the current environment, it is reproducible: if you run it twice,
+  the result will be the same<Hovernote>So much so that, as an optimization,
+  `garn` and `nix` store the result of the first run to avoid rerunning a second
+  time.</Hovernote>. And if you run it on different computers, then, with some
+  caveats, the result will be the same<Hovernote>So much so that, as an optimization,
+  if you run it a CI like <a href="https://garnix.io">garnix</a> and set it up
```
soenkehahn, line 54:
```suggestion
  if you run it on a CI like <a href="https://garnix.io">garnix</a> and set it up
```

website/src/docs/concepts.mdx
```diff
@@ -4,5 +4,119 @@ export const info = {
...
+A *check* is quite similar to a package, in that it also runs a command in a
+sandbox-like environment. But here we *do not* care about built files, but *only*
+the exit code<Hovernote>And the stdout and stderr of the process for debugging the check.</Hovernote>.
+
+You can run all checks in a project with "`garn check <PROJECT>`"
+
+## Executable
+
+An executable is a thing that can be run, without reference to the context in
+which it is run. In practice, they are run in the current environment via the
```
soenkehahn, line 88: > [...] without reference to the context in which it is run.

What do you mean by this?

website/src/docs/concepts.mdx
```diff
@@ -4,5 +4,119 @@ export const info = {
...
+"garn run" command. This provides a very easy way of making tools available
+to collaborators in your project.
+
+Usually executables will be a particular file in a package.
```
soenkehahn, line 92: Our `Executables` are often commands, including command line arguments. Under the hood they're translated into an executable file in a nix package, but that is not exposed through something of type `Package`.

website/src/docs/concepts.mdx
```diff
@@ -4,5 +4,119 @@ export const info = {
...
+
+A *project* is a collection of environments, packages, checks, and executables.
+The sole purpose of a project is to organize functionality in a coherent way.
+
+In the Typescript libraries, this is represented as an object with string keys
+and values of type `Environment | Check | Package | Executable`. You'll
+generally want to use language-specific helper functions (or, at a lower level,
+`mkProject`) in order to create a project.
+
+Projects can have *default* environments and executables. What this mean is
```
soenkehahn, line 104:
```suggestion
Projects can have *default* environments and executables. What this means is
```

website/src/docs/concepts.mdx
jkarni: I do think the current environment can be given as something of type Environment! We just can’t let it “escape” certain scopes, so it’s always as a sort of continuation rather than as a top-level value. 

This actually allows us to eg write a single function that generalizes both check and shell: it’d get passed two Environment arguments, one for the defaultEnvironment and one for the current environment. If you want what’s currently a check, you just ignore the second argument. Otherwise, you combine both environments.

website/src/docs/concepts.mdx
jkarni: That an Executable does not include an Environment. (Builds and Checks otoh do.)

website/src/docs/concepts.mdx
jkarni: Yeah, maybe my definition isn’t great. How would you define an Executable? Just like Build/Check, but always running in an augmentation of the current environment?

website/src/docs/concepts.mdx
soenkehahn: Hmm, but an `Executable` *does* include an `Environment`. We have a `.shell` function on `Environment` that creates `Executable`s.

jkarni: Yeah, talking to @alexdavid it became clear that the distinction between Environment and Executable isn’t an easy one. An executable can set environment variables etc, and can call provided programs. 

Not sure how to describe the differences then.

website/src/docs/concepts.mdx
soenkehahn: Yeah, something like that. Here's my attempt at a concepts page: #258. So that's how I would explain all the concepts. (I thought it was easiest to just start from scratch in another draft PR. Not saying we should abandon this one.)

PR #260, opened 2023-10-18
Support top-level executables

### LINKED ISSUE #254: Allow top-level Executables
This will allow users to add e.g. migration scripts easily:

```typescript
export const backend = mkGoProject({
  src = ".",
  ...
});


export const migrate: Executable = backend.shell`go run ./foo/migrate.go`
```

This would allow to run `garn run migrate`.

Part of this ticket should probably be to add something to the go example.

soenkehahn:
This PR implements support for `garn run`ning top-level `Executables`. This was mainly motivated by the go example where it was cumbersome to add some non-default executable like `go run scripts/migrate.go`.

Fixes #254.

Possibly easier to review commit by commit.

### DIFF
--- a/ts/process_compose.ts
+++ b/ts/process_compose.ts
@@ -31,5 +31,7 @@ export const processCompose = (
     nixRaw`pkgs.writeText "process-compose.yml" (builtins.toJSON ${processComposeConfig})`
   );
 
-  return emptyEnvironment.shell`${nixRaw`pkgs.process-compose`}/bin/process-compose up -f ${configYml}`;
+  const result = emptyEnvironment.shell`${nixRaw`pkgs.process-compose`}/bin/process-compose up -f ${configYml}`;
+  result.description = `processCompose(${Object.keys(executables).join(", ")})`;
+  return result;
 };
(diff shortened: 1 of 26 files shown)


ts/process_compose.ts
```diff
@@ -30,12 +31,5 @@ export const processCompose = (executables: Record<string, Executable>) => {
...
 
-  return mkProject(
-    {
-      description: "",
-      defaultEnvironment: emptyEnvironment,
-      defaultExecutable: emptyEnvironment.shell`${nixRaw`pkgs.process-compose`}/bin/process-compose up -f ${configYml}`,
-    },
-    {}
-  );
+  return emptyEnvironment.shell`${nixRaw`pkgs.process-compose`}/bin/process-compose up -f ${configYml}`;
```
alexdavid, line 34: Can we check that the description doesn't blow up here? I could imagine that the `configYml` being splatted in causes the description to be huge (low priority)

alexdavid (approved):
This looks great! So excited to have these!

ts/process_compose.ts
soenkehahn: Done in c09db3e3.

PR #258, opened 2023-10-18
Add draft of a concepts page

soenkehahn:
Fixes #166.

PR #262, opened 2023-10-19
Improve deno docs

### LINKED ISSUE #162: Add `jsdoc` comments

alexdavid:
Adds JSDoc comments and other changes to improve the automated docs at docs.deno.land.

Reviewing one commit at a time is recommended.

Resolves #162


ts/base.ts
```diff
@@ -1,7 +1,3 @@
-export type ShouldNotRun =
-  | { tag: "ShouldNotRun" }
-  | { tag: "UnexpectedError"; reason: string };
-
```
alexdavid, line 4: This type was not used anywhere

ts/check.ts
```diff
@@ -1,9 +1,22 @@
...
+ * `Check`s are commands (usually shell commands) that are used to define
+ * reproducible automated tests or other checks.
+ *
+ * `Check`s are run in a sandbox. This means they don't have access to your
+ * locally installed tools, local file system, or the internet. Instead, they
+ * are always based on an underlying `Environment`, and through that will have
+ * access to the tools that the environment provides
+ *
+ * You can run `Check`s with `garn check`.
+ */
```
alexdavid, line 14: I copied a lot of these from the #258 draft. Overall I think this duplication is nice since these concepts are so important that it is nice to have it show up not only on the website, but also in the deno docs and in editors with typescript LSP.

ts/environment.ts
```diff
@@ -3,59 +3,118 @@ import { hasTag, nixSource } from "./internal/utils.ts";
...
+ * variables that need to be set.
+ *
+ * You can enter an `Environment` with `garn enter`.
+ */
 export type Environment = {
   tag: "environment";
   nixExpression: NixExpression;
+  /**
+   * Creates a new environment based on this one that includes the specified nix packages.
+   */
```
alexdavid, line 28: As mentioned in #108 deno doc does not support rendering these yet unfortunately. However I think the fact that they show up on LSP hover is useful still.

ts/project.ts
```diff
@@ -67,39 +68,36 @@ export function isProject(p: unknown): p is Project {
 
 type Nestable = Environment | Package | Executable | Check;
 
-type ProjectSettings = {
-  description: string;
-  defaultEnvironment?: Environment;
-  defaultExecutable?: Executable;
-};
```
alexdavid, line 74: We either need to inline this or export it through `mod.ts` in order to get visibility of the options here in docs.deno.land.

I opted for inlining for now since it wasn't used anywhere else.

soenkehahn (approved):
Left a bunch of comments, but this is very nice!

ts/environment.ts
```diff
@@ -3,59 +3,118 @@ import { hasTag, nixSource } from "./internal/utils.ts";
...
-export const check = (
+/**
+ * Creates a new shell script `Check`, run in the `emptyEnvironment`.
+ *
+ * Example:
+ * ```typescript
+ * // Will fail if any `TODO`s are found in the build artifacts of myPkg:
+ * garn.check`! grep -r TODO ${myPkg}`;
+ * ```
+ */
```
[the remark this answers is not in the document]
alexdavid: So the example here I think shows that you can check a package. Whether the package has source files or not is up to whatever created the package. That being said, maybe packgaes of source files are unidiomatic for garn. If this is the case I can update the example.

I think having a global check might still be nice since maybe I want to write a check that checks that a binary runs successfully in the empty environment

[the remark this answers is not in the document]
soenkehahn: Oh, I missed the `${myPkg}`. Yeah, LGTM for now then.

PR #263, opened 2023-10-19
Add website checks to top-level checks

soenkehahn:


flake.nix
```diff
@@ -151,6 +156,14 @@
                   --replace !/usr/bin/env !${pkgs.coreutils}/bin/env
                 just ${recipe}
               '';
+            websiteOutputs = ((import ./website/flake.nix).outputs {
+              self = websiteOutputs;
+              inherit nixpkgs flake-utils gomod2nix-repo npmlock2nix-repo;
+            });
```
soenkehahn, line 162: Is there a better way to do this without having to manually copy the inputs of `website/flake.nix` into this top-level flake?

I skimmed https://github.com/NixOS/nix/issues/3978 and tried out a few solutions mentioned in there, but none are really nice.

cc: @R-VdP @jfroche

soenkehahn: I think I solved it with https://github.com/divnix/call-flake.

alexdavid (approved):

flake.nix
R-VdP: Probably best to float the call-flake call out so that it's only evaluated once. Nix is not very advanced when it comes to common subexpression elimination.

soenkehahn: I found where this is extracted from: https://github.com/NixOS/nix/blob/4d17c59d8d059a5b39f1d1da2b58f2ec8da44861/src/libexpr/flake/call-flake.nix

PR #264, opened 2023-10-19
Remove moduleName from mkGoProject

### LINKED ISSUE #220: Remove module name from mkGoProject

soenkehahn:
Fixes #220.

I also bumped the ts lib version, since this is a breaking change.

This means that we can't `garn run` the projects that we get from `mkGoProject`
anymore. But we have ways of providing `Executables` for that.

### DIFF
--- a/examples/monorepo/garn.ts
+++ b/examples/monorepo/garn.ts
@@ -2,17 +2,17 @@ import * as garn from "http://localhost:8777/mod.ts";
 
 export const backend = garn.go.mkGoProject({
   description: "example backend server in go",
-  moduleName: "go-http-backend",
   src: "backend",
 });
 
-export const haskell = garn.haskell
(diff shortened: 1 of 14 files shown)


examples/monorepo/garn.ts
```diff
@@ -21,7 +21,7 @@ export const npmFrontend = garn.javascript.mkNpmProject({
 });
 
 export const startAll = garn.processCompose({
-  backend: backend.defaultExecutable!,
```
alexdavid, line 24: I'm glad these `!`s are finally out of our examples

alexdavid (approved):
This looks good. I think there is still a reference in the website asciinemas, but I think it's not a huge priority in getting that updated since it's not really designed to be a reference on the API. I can re-record the asciinemas next week.

PR #265, opened 2023-10-19
Implement `.build` to create `Package`s

### LINKED ISSUE #255: [npm] Allow to add a `Package` that contains the bundled code
This has to be pretty flexible, because people use different bundlers with different command line flags (or no bundler at all for e.g. node backend projects). So maybe an API like this makes sense:

```typescript
export const frontend = mkNpmProject({
  src = ".",
  description: "An NPM frontend",
  nodeVersion: "18",
});

export const bundle: Package =
  frontend.package`
    webpack src --or-whatever
    cp -rv dist/* $out/
`;
```

Which would allow `garn build bundle`.

soenkehahn:
Fixes #255.

### DIFF
--- a/justfile
+++ b/justfile
@@ -122,6 +122,9 @@ check-examples:
   just run-garn haskell check
   just run-garn haskell run helloFromHaskell
   just run-garn frontend-create-react-app check
+  rm examples/frontend-create-react-app/result
+  just run-garn frontend-create-react-app build bundle
+  ls examples/frontend-create-react-app/result/index.html
   just run-garn frontend-yarn-webpack check
   just run-garn go-http-backend check
   just run-garn monorepo check
--- a/ts/environment.ts
+++ b/ts/environment.ts
@@ -1,7 +1,6 @@
-import { Package } from "./package.ts";
-import { hasTag, nixSource } from "./internal/utils.ts";
 import { Check } from "./check.ts";
 import { Executable } from "./executable.ts";
+import { hasTag, nixSource } from "./internal/utils.ts";
 import {
   NixStrLitInterpolatable,
   NixExpression,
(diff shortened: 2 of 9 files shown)


ts/environment.ts
```diff
@@ -28,12 +29,18 @@ export const check = (
   ...args: Array<Interpolatable>
 ) => emptyEnvironment.check(s, ...args);
 
+export const build = (
+  s: TemplateStringsArray,
+  ...args: Array<Interpolatable>
+) => emptyEnvironment.build(s, ...args);
```
alexdavid, line 35: I don't think this will conflict with #262, but if that makes it in first we should switch this to `function` and add a jsdoc:

```suggestion
/**
 * Creates a new shell script `Package`, run in the `emptyEnvironment`.
 *
 * Copy any artifacts that you want to keep from the build in `$out`.
 *
 * Example:
 * ```typescript
 * import * as pkgs from "https://garn.io/ts/v0.0.9/nixpkgs.ts";
 * // Creates a package containing a 4096-bit RSA ssh key
 * garn.build`
 *   ${pkgs.openssh}/bin/ssh-keygen -b 4096 -t rsa -N "" -f $out/id_rsa
 * `;
 * ```
 */
${pkgs.openssh}/bin/ssh-keygen -b 4096 -t rsa -N "" -f $out/id_rsa
export function build(
  s: TemplateStringsArray,
  ...args: Array<Interpolatable>
) {
  return emptyEnvironment.build(s, ...args);
}
```

Or maybe if the ssh example is too complex, maybe just soemthing simple like
```
    ${pkgs.cowsay}/bin/cowsay moo > $out/moo
```

justfile
```diff
@@ -122,6 +122,9 @@ check-examples:
   just run-garn haskell check
   just run-garn haskell run helloFromHaskell
   just run-garn frontend-create-react-app check
+  rm examples/frontend-create-react-app/result
+  just run-garn frontend-create-react-app build bundle
+  ls examples/frontend-create-react-app/result/index.html
```
alexdavid, line 127: I like this way of checking the build command end to end :slightly_smiling_face:

alexdavid (approved):
This is so cool!

PR #266, opened 2023-10-20
Prevent exporting addCheck directly

### LINKED ISSUE #223: Exporting foo.addCheck("check") silently fails
`addCheck("blah")` doesn't return a Project, but instead a template literal that returns a project. So when you forget to (or don't know that you should) call the template literal, you no longer get the project that you thought you had, without any warnings or explanations.

alexdavid:
Resolves #223

### DIFF
--- a/src/Garn.hs
+++ b/src/Garn.hs
@@ -2,24 +2,23 @@
 
 module Garn
   ( Env (..),
-    run,
     readOptionsAndConfig,
     runWith,
+    run,
   )
 where
 
+import Control.Exception (catch)
 import Control.Monad (forM_, when)
 import Development.Shake (Exit (Exit), cmd)
 import Garn.Common (currentSystem, nixArgs)
+import qualified Garn.Errors
 import Garn.GarnConfig
 import Garn.Init
 import Garn.Optparse
-import Paths_garn
 import System.Directory (doesFileExist)
 import System.Exit (ExitCode (..), exitWith)
 import System.IO (Handle, hPutStrLn, stderr)
-import qualified System.IO
-import qualified System.Posix.User as POSIX
 import System.Process
 
 data Env = Env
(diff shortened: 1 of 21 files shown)


src/Garn.hs
```diff
@@ -2,24 +2,23 @@
 
 module Garn
   ( Env (..),
-    run,
     readOptionsAndConfig,
     runWith,
+    run,
   )
```
soenkehahn, line 8: I think we might be able to export less items here, since we're not using them in tests anymore.

src/Garn.hs
```diff
@@ -28,11 +27,12 @@ data Env = Env
...
-run :: IO ()
-run = do
-  env <- productionEnv
-  options <- readOptionsAndConfig
-  runWith env options
+run :: Env -> IO ()
+run env =
+  (readOptionsAndConfig >>= runWith env)
+    `catch` \(Garn.Errors.UserError err) -> do
+      hPutStrLn stderr $ "[garn] " <> err
```
soenkehahn, line 34:
```suggestion
      hPutStrLn stderr $ "[garn] Error: " <> err
```

To distinguish this from only informational output. TIOLI.

ts/internal/may_not_export.ts
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * The reason for this is to catch potential mistakes like the following:
+ * ```typescript
+ * export const myProject = getProjectSomehow().addCheck("my-test")
+ * ```
+ *
+ * The above example is missing the call on the tag template literal function
+ * returned by `addCheck`.
+ */
+export function markAsMayNotExprt(
```
soenkehahn, line 18:
```suggestion
export function markAsMayNotExport(
```

ts/internal/may_not_export.ts
```diff
@@ -0,0 +1,42 @@
...
+  // deno-lint-ignore ban-types
+  value: object,
+  reason: (exportName: string) => string
+) {
+  // @ts-expect-error - SAFETY: typescript does not allow setting
+  // `MAY_NOT_EXPORT` on T here. However, it is safe to set arbitrary keys on
+  // non-primative types which `value: object` enforces
```
soenkehahn, line 25:
```suggestion
  // non-primative types, which `value: object` enforces.
```

ts/internal/may_not_export.ts
```diff
@@ -0,0 +1,42 @@
...
+  value[MAY_NOT_EXPORT] = reason;
+}
+
+/**
+ * Throws a runtime error if the passed value has been marked as may not export
+ * using `markAsMayNotExprt`.
```
soenkehahn, line 31:
```suggestion
 * using `markAsMayNotExport`.
```

soenkehahn (approved):
Left some comments. Otherwise LGTM.

Also, since this is a breaking change, should it bump the ts lib version?

(There's also #264 in flight. So -- if that gets in first -- this has to bump based on that.)

soenkehahn (approved):

PR #267, opened 2023-10-20
Implement addExecutable

alexdavid:
This implements `addExecutable` on project, and allows project-nested executables to be run with `/` for now. I'm not sure if it's possible to even invoke flake apps whose name contains a `.`. I experimented with `%2e` which seems to get past one layer of parsing but fails elsewhere. In the future we'll need to do something smarter with haskell to print `.` in the help text and convert it to some other character to invoke the flake app, but for now this seems more useful than not having it.

### DIFF
--- a/test/spec/RunSpec.hs
+++ b/test/spec/RunSpec.hs
@@ -60,6 +60,20 @@ spec =
           stdout output `shouldBe` "Hello, world!\n"
           exitCode output `shouldBe` ExitSuccess
 
+        it "runs manually specified exeutables" $ do
+          writeFile
+            "garn.ts"
+            [i|
+              import * as garn from "#{repoDir}/ts/mod.ts"
+              export const myEnv = garn.mkProject({
+                description: "my project",
+                defaultEnvironment: garn.emptyEnvironment,
+              }, {}).addExecutable("hello")`echo Hello, world!`;
+            |]
+          output <- runGarn ["run", "myEnv/hello"] "" repoDir Nothing
+          stdout output `shouldBe` "Hello, world!\n"
+          exitCode output `shouldBe` ExitSuccess
+
         it "allows specifying argv to the executable" $ do
           writeFile
             "garn.ts"
--- a/test/spec/RunSpec.hs
+++ b/test/spec/RunSpec.hs
@@ -7,7 +7,6 @@ import Data.List (sort)
 import Data.Maybe (isJust)
 import Data.String.Interpolate (i)
 import Data.String.Interpolate.Util (unindent)
-import Development.Shake (cmd_)
 import System.Directory
 import System.Environment (lookupEnv)
 import System.Exit (ExitCode (..))
@@ -70,7 +69,7 @@ spec =
                 defaultEnvironment: garn.emptyEnvironment,
               }, {}).addExecutable("hello")`echo Hello, world!`;
             |]
-          output <- runGarn ["run", "project/hello"] "" repoDir Nothing
+          output <- runGarn ["run", "project.hello"] "" repoDir Nothing
           stdout output `shouldBe` "Hello, world!\n"
           exitCode output `shouldBe` ExitSuccess
 
@@ -90,7 +89,7 @@ spec =
                 description: "b",
               }, { b });
             |]
-          output <- runGarn ["run", "c/b/a"] "" repoDir Nothing
+          output <- runGarn ["run", "c.b.a"] "" repoDir Nothing
           stdout output `shouldBe` "executable in a\n"
           exitCode output `shouldBe` ExitSuccess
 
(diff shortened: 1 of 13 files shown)


ts/internal/runner.ts
```diff
@@ -155,6 +155,11 @@ const findExportables = (
   for (const [name, value] of Object.entries(config)) {
     if (isProject(value)) {
       result[name] = value;
+      const nested = findExportables(value);
+      for (const key in nested) {
+        if (key === "defaultExecutable") continue;
+        result[`${name}/${key}`] = nested[key];
+      }
```
soenkehahn, line 162: What will this code do for `Project`s within `Project`s? I'm not sure it'll do something sensible. So maybe filter out `Project`s here? At least until we have tests for nesting `Project`s?

soenkehahn (changes requested):

ts/project.ts
```diff
@@ -143,17 +162,36 @@ const proxyEnvironmentHelpers = () => ({
...
     }
     return (s: TemplateStringsArray, ...args: Array<string>) => {
-      const newCheck = this.check(s, ...args);
+      const newCheck = { [name]: this.check(s, ...args) } as {
+        [n in Name]: Check;
+      };
       return {
         ...this,
-        [name]: newCheck,
+        ...newCheck,
```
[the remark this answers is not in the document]
alexdavid: This changed because of the need for the `as` cast above (see my next comment)

ts/project.ts
```diff
@@ -143,17 +162,36 @@ const proxyEnvironmentHelpers = () => ({
...
     if (this.defaultEnvironment == null) {
       throw new Error(
         `'.addCheck' can only be called on projects with a default environment`
       );
     }
     return (s: TemplateStringsArray, ...args: Array<string>) => {
-      const newCheck = this.check(s, ...args);
+      const newCheck = { [name]: this.check(s, ...args) } as {
+        [n in Name]: Check;
+      };
```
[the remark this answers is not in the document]
alexdavid: The `as` cast is a bit sad, but luckily typescript will error if the type isn't close (e.g. we expect a `{ [Name]: Project }` instead of `{ [Name]: Check }` here would require an `as any as ...`.

I'm not quite sure why typescript was losing the type here, but while `name` is correctly typed as `Name`, `{ [name]: Check }` was becoming into `{ [string]: Check }`. I might keep looking into this this weekend to figure out why this is happening

ts/internal/runner.ts
alexdavid: I haven't tried it, but I would imagine it would properly nest as expected, no? `findExportables` currently just creats a map from string to Projects/executables.

So in that case a project's keys would be concatenated here, right?

alexdavid: Added a test for this, but to make it pass we needed to add `Project` to `Nestable`. I think this might be the right time to add it, but also if we're not ready to add it yet then I think maybe it's fine to leave this as is without the project check since typescript will prevent nested projects from occurring anyway

ts/project.ts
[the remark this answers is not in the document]
soenkehahn: Thanks for explaining. Yeah, typescript sometimes assumes type `string` and often that's whet we want, but sometimes not...

ts/internal/runner.ts
soenkehahn: Yeah, I think it's fine to have nested Executables now.

test/spec/RunSpec.hs
```diff
@@ -60,6 +60,40 @@ spec =
...
+                defaultExecutable: garn.shell`echo executable in a`,
+              }, {});
+              const b = garn.mkProject({
+                description: "b",
+              }, { a });
+              export const c = garn.mkProject({
+                description: "b",
+              }, { b });
+            |]
+          output <- runGarn ["run", "c/b/a"] "" repoDir Nothing
```
jkarni, line 93: Can't we just replace dots with slashes, so the *interface* still uses dots? (It doesn't matter *that* much that the flake file is different)

ts/project.ts
```diff
@@ -143,17 +162,36 @@ const proxyEnvironmentHelpers = () => ({
     return this.defaultEnvironment.build(s, ...args);
   },
 
+  addExecutable<T extends Project, Name extends string>(this: T, name: Name) {
+    if (this.defaultEnvironment == null) {
+      throw new Error(
+        `'.addExecutable' can only be called on projects with a default environment`
+      );
+    }
+    return (s: TemplateStringsArray, ...args: Array<string>) => {
```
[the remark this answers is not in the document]
jkarni: Done.

jkarni:
In fact nix doesn't seem to allow running of apps with dots in their names. So I created a small translation layer between the user-facing names (which now have `.`) and the nix ones (which still have `/`).

@soenkehahn could you re-review?

soenkehahn (approved):
Nice!

PR #268, opened 2023-10-20
Tweak deploy recipes

soenkehahn:
This modifies the two deploy recipes that we have.
They now

1. Don't do anything in your local checkout.
   Instead they both take a commit and clone a fresh, clean copy into a temporary directory.
2. They don't push anything.
   Instead they print out the temporary directory for you to inspect and push manually.

I think these recipes could be tweaked further, but I do think this is an improvement.

### DIFF
--- a/justfile
+++ b/justfile
@@ -15,40 +15,52 @@ fmt: fmt-nix fmt-haskell fmt-typescript
 check: fmt-nix-check fmt-haskell-check hpack-check fmt-typescript-check
 
 # Deploy the current website
-deploy-website: build-install-script
+deploy-website commit: build-install-script
   #!/usr/bin/env bash
   set -eux
-  if [[ -z $(git status -s) ]]; then
-    COMMIT=$(git rev-parse HEAD)
-    TMP_DIR=$(mktemp --directory)
-    (cd website && npm install && npm run build)
-    ls website/dist/*
-    mv website/dist/* $TMP_DIR
-    git checkout gh-pages
-    cp -rv $TMP_DIR/* .
-    rm -rf $TMP_DIR
-    git add .
-    git commit -m "Website update from commit $COMMIT"
-    echo "Created a new commit. It is not yet pushed."
-  else
-    echo "Working directory is dirty. Please commit, stash or reset before continuing"
-  fi
+
+  TMP_DIR=$(mktemp --directory)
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -15,40 +15,52 @@ fmt: fmt-nix fmt-haskell fmt-typescript
...
+
+  TMP_DIR=$(mktemp --directory)
+  DIST_DIR=$TMP_DIR/dist
+  mkdir -p $DIST_DIR
+  cd $TMP_DIR
+  git clone git@github.com:garnix-io/garn
+  cd garn
+  git reset --hard {{ commit }}
+  COMMIT=$(git rev-parse HEAD)
+  (cd website && npm install && npm run build)
```
alexdavid, line 30: Since this is in the root now and we don't have npm/node in our root flake file this may fail. Should we add them? (This will be so much nicer when we are able to migrate the entire garn repo to use garn :slightly_smiling_face:)

justfile
```diff
@@ -15,40 +15,52 @@ fmt: fmt-nix fmt-haskell fmt-typescript
...
+  git reset --hard {{ commit }}
+  git tag $tag
+
+  just codegen
+
+  git checkout gh-pages
+  mkdir -p tmp-$tag
+  git --work-tree=tmp-$tag checkout {{ commit }} -- ts
+  mv tmp-$tag/ts ts/$tag
+  mv ts/nixpkgs.ts ts/$tag
```
alexdavid, line 58: We used to also commit the latest release directly under `ts`, maybe this is an anti-pattern that we don't want to encourage. If this is the case maybe we should remove the latest release on the current `gh-pages`?

alexdavid (approved):
I like that these feel better that a stray file in our working tree won't accidentally get deployed

justfile
soenkehahn: I was really confused by this. I *did* find the those files on the `gh-pages` branch, but where in the old script were we doing that?

justfile
soenkehahn: Yeah, we should change this, but this was the same before this PR. I didn't want to change too much at once.

justfile
soenkehahn: We talked about this offline. I'm removing the unversioned library files in #276.

PR #271, opened 2023-10-21
Add (failing) test for passing command line flags to `Executable`s

soenkehahn:


src/Garn/Optparse.hs
```diff
@@ -11,16 +12,44 @@ module Garn.Optparse
...
+      error "BUG: parser shouldn't include any leftover args for run commands!"
+    Right (options, []) -> pure options
+    Right (_options, _ : _) -> do
+      -- This is a parse error, since only `run` commands can have arbitrary
+      -- leftover arguments. To get the normal error behavior, we parse the
+      -- flags again, this time with `execParserPure`, which processes all
+      -- arguments.
+      handleParseResult $ execParserPure parserPrefs parserInfo allArgs
+    Left err ->
+      handleParseResult $ Failure $ parserFailure parserPrefs parserInfo err contexts
```
soenkehahn, line 52: This has gotten much more complicated than I thought it would be. But I haven't found an easier way to do this with `optparse-applicative`.

test/spec/Garn/OptparseSpec.hs
```diff
@@ -74,6 +74,32 @@ spec = around_ (hSilence [stderr]) $ do
         command <- testWithGarnTs ["run", "project", "more", "args"] ("project" ~> targetConfig)
         command `shouldBe` Run (CommandOptions "project" targetConfig) ["more", "args"]
 
+      it "parses run commands with additional flags (starting with `-`)" $ do
```
jkarni, line 77: So you need — only for —help to be parsed by the inner command? That feels unintuitive to me. Should we maybe require it for all arguments? I think not needing it made sense when we didn’t have the past command (eg “foo” in “garn build foo”) be optional, but doesn’t anymore.

jkarni (commented):
Passing *flags* (rather than string arguments) to the subprocess is already possible with `--`; as in `garn run foo bar -- --baz`. It's not a great system, but I think requiring `--` only when `run` happens to have a conflicting option isn't great either (both because it's not clear from the command alone who is getting the option, and because of backwards compatibility).

jkarni:
For whatever reason, I need to pass `--` *before* the target: `garn run -- dev -f` works, but `garn run dev -- -f` does not.

Parsing `--` outside of optparse-applicative sounds like a good idea!

PR #278, opened 2023-10-21
Implement editGarnConfig to spin up a pre-configured vscodium

### LINKED ISSUE #194: garn edit
It's quite annoying and non-obvious to set up your text editor correctly in order to get LSP/IDE goodness for garn.ts files. 

We could help out somehow. For instance providing a `garn.edit` Project that has an executable that is vscodium with the appropriate settings, so users could run `garn run edit`

soenkehahn:
Fixes #194.

### DIFF
--- a/ts/edit.ts
+++ b/ts/edit.ts
@@ -39,8 +39,8 @@ const settingsFile = (): NixExpression => {
 };
 
 /**
- * A preconfigure state.vscdb sqlite file that will prevent vscodium to open a
- * welcome tab for the deno extension.
+ * A preconfigured state.vscdb sqlite file that will prevent vscodium from
+ * opening a welcome tab for the deno extension.
  */
 const stateFile = (): Package => {
   const sqliteScript = outdent`
@@ -61,8 +61,8 @@ const stateFile = (): Package => {
 };
 
 /**
- * This is a fake home directory that contains preconfigured vscodium settings files,
- * to have a nice experience when editing garn.ts files.
+ * This is a fake home directory that contains preconfigured vscodium settings
+ * files that provide a nicer experience when editing garn.ts files.
  */
 const configDir = (): Package => {
   return garn.build`
@@ -105,6 +105,8 @@ export const editGarnConfig: garn.Executable = (() => {
     fi
 
     export HOME=$TEMP_DIR
+    export XDG_CONFIG_HOME=$TEMP_DIR/.config
+    export XDG_CACHE_HOME=$TEMP_DIR/.cache
 
     ${vscodium}/bin/codium --new-window --disable-workspace-trust ./garn.ts --wait
 
--- a/ts/edit.ts
+++ b/ts/edit.ts
@@ -1,0 +1,115 @@
+import outdent from "https://deno.land/x/outdent@v0.8.0/mod.ts";
+import * as garn from "./mod.ts";
+import { Package } from "./mod.ts";
+import { NixExpression, nixRaw } from "./nix.ts";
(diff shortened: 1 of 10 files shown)


jkarni (changes requested):
Cool! It isn't really working very well for me, though. I manage to start up the editor, but I get "Cannot find module 'http://localhost:8777/mod.ts' or its corresponding type declarations.ts(2307)" on startup (same with `garn.io` domain.

ts/edit.ts
```diff
@@ -0,0 +1,115 @@
...
+      "editor.defaultFormatter": "denoland.vscode-deno",
+    },
+    "update.mode": "none",
+  };
+  return writeTextFile("vscodium-config", JSON.stringify(settings, null, 2));
+};
+
+/**
+ * A preconfigure state.vscdb sqlite file that will prevent vscodium to open a
+ * welcome tab for the deno extension.
```
jkarni, line 43: I got the welcome tab.

soenkehahn:
@jkarni: Hmm, for some reason this doesn't seem to work as expected on your machine. Both the deno caching and putting the settings in place. It works on all the machines I've tested on (two linux, one mac). Maybe we can debug this together?

website/src/docs/getting_started.mdx
```diff
@@ -90,7 +90,24 @@ without having to install any language-specific tools or learn how to use them.
...
 `garn.ts` files are powered by [deno](https://deno.com/).
 A lot of the convenience and power for editing your `garn.ts` files comes
-from having a working Deno LSP. Installing `deno` is easy with `garn` itself:
+from having a working Deno LSP.
+
+### `editGarnConfig`
+
+`garn` offers an `Executable` `editGarnConfig` that you can add to your
+`garn.ts` file and that you can run to spin up a vscodium editor that is
+pre-configured for editing `garn.ts` files. This is the easiest way to get a
```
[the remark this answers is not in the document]
soenkehahn: I reworked this section in 55bec1c0. What do you think?

Regarding both macs: I tested this only on an arm mac. I propose to ignore intel macs for the time being (until someone complains). We don't have a binary cache for intel macs. So even if everything works *correctly*, everything will be terribly slow, including installing `garn` and building any project. (And testing this editor.) I think -- if we decide intel macs are important enough to support well -- the next step is to set up a builder on garnix.io for that. In the meantime we should add a warning to the website that this doesn't work well on intel macs.

website/src/docs/getting_started.mdx
```diff
@@ -17,32 +17,37 @@ The easiest way to install <Garn/> is with the following command:
...
-  ```
+
+```
 extra-substituters = https://cache.garnix.io
 extra-trusted-public-keys = cache.garnix.io:CTFPyKSLcx5RMJKfLo5EEPUObbA78b0YQ2DTCJXqr9g=
-  ```
-  To `/etc/nix/nix.conf` (note that you'll need sudo).
+```
+
+To `/etc/nix/nix.conf` (note that you'll need sudo).
```
jkarni, line 35: Maybe doesn't need to be now, but just occurred to me that we need to tell users to restart the daemon after this change.

jkarni (approved):

website/src/docs/getting_started.mdx
soenkehahn: Good point! I think we don't though. At least `nix show-config` shows the the cache without restarting. Although we should make sure that the cache is actually being used.

soenkehahn: Yeah, never mind. `nix show-config` is just lying.

PR #309, opened 2023-10-26
Implement --version flag

### LINKED ISSUE #207: Implement `--version` flag

soenkehahn:
Fixes #207.

### DIFF
--- a/justfile
+++ b/justfile
@@ -90,10 +90,10 @@ ormolu mode:
   fi
 
 fmt-typescript:
-  prettier --write $(fd .ts ts | grep -v nixpkgs.ts)
+  prettier --write $(fd .ts ts | grep -v 'nixpkgs.ts\|version.ts')
 
 fmt-typescript-check:
-  prettier --check $(fd .ts ts | grep -v nixpkgs.ts)
+  prettier --check $(fd .ts ts | grep -v 'nixpkgs.ts\|version.ts')
 
 hpack:
   hpack
@@ -151,7 +151,7 @@ codegen: hpack && typescript-check
   cabal run codegen
 
 typescript-check *args="":
-  deno check ts/*.ts {{ args }}
+  deno check ts/*.ts ts/**/*.ts {{ args }}
   echo checked!
 
 check-isolated-garn:
--- a/package.yaml
+++ b/package.yaml
@@ -8,6 +8,7 @@ copyright: "AllRightsReserved"
 verbatim:
   data-files: |
     ts/**/*.ts
+    ts/**/*.json
   cabal-version: 2.4
 
 default-extensions:
--- a/package.yaml
+++ b/package.yaml
@@ -55,6 +55,7 @@ dependencies:
   - process
   - shake
   - string-conversions
+  - template-haskell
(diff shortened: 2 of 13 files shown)


test/spec/GarnSpec.hs
```diff
@@ -60,67 +61,69 @@ spec = do
...
-                type DenoOutput = {
-                  "some incompatible type": null,
-                  garnTsLibVersion: string
-                };
+      it "outputs a version with --version" $ onTestFailureLogger $ \onTestFailureLog -> do
+        output <- runGarn ["--version"] "" repoDir Nothing
+        onTestFailureLog output
+        stdout output `shouldBe` "v0.0.14\n"
+        stderr output `shouldBe` ""
+        exitCode output `shouldBe` ExitSuccess
```
soenkehahn, line 76: The diff for this file looks terrible. It's probably easier to review with hidden white space changes. Anyways, this is the only real now test.

ts/internal/version.ts
```diff
@@ -0,0 +1,4 @@
+// This is in its own file because the `with { type: "json" }` syntax breaks code formatting.
+import version from "./version.json" with { type: "json" };
```
alexdavid, line 2: Should we consider just reading this with `Deno.readFileSync` instead if it causes issues with the formatter?

alexdavid (approved):
Looks good

soenkehahn:
I realized (prompted by @alexdavid's comment) that this PR probably breaks `garn init` currently, since we don't ship that `version.json` file in the cabal install.

package.yaml
```diff
@@ -8,6 +8,7 @@ copyright: "AllRightsReserved"
 verbatim:
   data-files: |
     ts/**/*.ts
+    ts/**/*.json
```
soenkehahn, line 11: This is using cabal globs verbatim and that doesn't allow file globs without extensions, see https://cabal.readthedocs.io/en/stable/cabal-package.html#pkg-field-data-files.

justfile
```diff
@@ -151,7 +151,7 @@ codegen: hpack && typescript-check
   cabal run codegen
 
 typescript-check *args="":
-  deno check ts/*.ts {{ args }}
+  deno check ts/*.ts ts/**/*.ts {{ args }}
```
soenkehahn, line 154: Unrelated to this PR, but before we didn't typecheck files in subdirs.

ts/internal/version.ts
soenkehahn: That won't work in the case when the ts lib is imported from `https://garn.io/ts/...`. Also, it'd be hard to figure out the path when reading this from the cabal data files. I think it's easier to rely on deno's importing mechanism here.

We *do* have to make sure that the `version.json` file gets

- deployed with `just release-ts` (I checked that manually),
- included in the cabal data files (see the new change in `package.yaml`), and
- served for local development with `just fileserver` (also tested manually).

alexdavid: Ah of course, forgot about the URL imports :confounded:

alexdavid (approved):

PR #312, opened 2023-10-27
Update nixpkgs (both garn's and for the generated nixpkgs.ts)

soenkehahn:
Currently we recommend in the getting started guide to put `deno` into a garner
environment to make it available for editing `garn.ts` files. However it's not
the same `deno` version, `garn` currently uses `1.34.1`, but in the generated
`nixpkgs.ts` it's `1.33.3`. After this PR it'll be `1.37.2` in both places.

I've updated the patch-level version of `ghc` since 9.4.7 is the default for
that branch and it seemed that more libraries where cached on cache.nixos.org
for that.

### DIFF
--- a/website/flake.nix
+++ b/website/flake.nix
@@ -1,5 +1,5 @@
 {
-  inputs.nixpkgs.url = "github:NixOS/nixpkgs/21443a102b1a2f037d02e1d22e3e0ffdda2dbff9";
+  inputs.nixpkgs.url = "github:NixOS/nixpkgs/6fc7203e423bbf1c8f84cccf1c4818d097612566";
   inputs.flake-utils.url = "github:numtide/flake-utils";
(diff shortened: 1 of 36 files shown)


website/flake.nix
```diff
@@ -329,6 +329,8 @@ COMMIT;";
     fi
 
     export HOME=\$TEMP_DIR
+    export XDG_CONFIG_HOME=\$TEMP_DIR/.config
+    export XDG_CACHE_HOME=\$TEMP_DIR/.cache
```
alexdavid, line 333: Where did this change come from? This flake file is generated by garn, right?

alexdavid (approved):
Looks good. Did the trailing commas get added because of the update to deno? I thought our `just fmt` wasn't using deno's formatter because my deno LSP always tries to disagree with the formatting.

website/flake.nix
soenkehahn: This was a change to the `editGarnConfig` `Executable` that we include in `website/garn.ts`. So right now this change also appears on `main` after running `cd website ; garn generate`.

Issue #317, opened 2023-10-27
Backwards- and forwards-compatibility plan


jkarni:
I just had a discussion with @soenkehahn about how we can ensure the smoothest possible way forward for breaking changes in the garn TS library <-> haskell executable. The idea (@soenkehahn, I'm just the scribe), roughly, is this.

1. The tags we release the TS library will also indicate what version of garn-the-executable is compatible with that library. We won't use semver or anything else: just the executable with tag vX.X.X is compatible with the library with tag vX.X.X. (Semver can still be meaningfully used for e.g. *within TS* versioning.)
2. garn-the-executable will try to parse the output of runner.ts. If it succeeds, no attention is paid to version.
3. If it fails, the TS lib will be providing it's own version number in the GarnConfig object, and the executable will look at that. And then *download the corresponding version of the executable* (which is, apparently, only 2 or so MB). After downloading it, it will run *that* executable against the library (and also, if the version is newer, do an automatic upgrade of the installed version?). Because of the nix cache, we only pay for the (already small) download cost once per version, or once prior to garbage collection.
4. This means: no worrying about backwards or forwards compatibility when releasing. The executable just has to work with the current version of the lib. 

In general I think this is very nice, if somewhat unorthodox. 

The task for this issue, then, is: download any tag of garn on-demand, and rerun that, if there's a version incompatiblity.

Issue #306, opened 2023-10-26
Add documentation about how to update `garn`


soenkehahn:
Something like

```bash
$ nix profile list
# look up which entry is `garn`
$ nix profile remove 42
$ nix profile install github:garnix-io/garn
```

Or is there a better way than this?

PR #320, opened 2023-10-27
Implement deno <-> haskell JSON version check

soenkehahn:
This implements a first step in the direction of #317 and should make error messages on json version mismatches much better. Once we have #306, the error message should include a link to that.

Since this is a breaking change to the json format this PR also bumps the ts lib version. But from this version onwards we should have better error messages. :)

### DIFF
--- a/src/Garn/GarnConfig.hs
+++ b/src/Garn/GarnConfig.hs
@@ -15,6 +15,7 @@ import Data.Aeson
     withObject,
     (.:),
   )
+import Data.List (intercalate)
 import Data.Map (Map)
 import Data.String (IsString (fromString))
 import Data.String.Interpolate (i)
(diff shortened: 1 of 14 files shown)


src/Garn/GarnConfig.hs
```diff
@@ -30,9 +31,24 @@ import System.IO.Temp (withSystemTempFile)
...
-  = Success GarnConfig
-  | UserError String
-  deriving (Eq, Show, Generic, FromJSON)
+  = Success Version GarnConfig
+  | UserError Version String
+  deriving stock (Eq, Show, Generic)
+
+instance FromJSON DenoOutput where
+  parseJSON = withObject "DenoOutput" $ \o -> do
+    garnTsLibVersion <- o .: fromString "garnTsLibVersion"
```
soenkehahn, line 40: Technically we're not using the `Version` field in `DenoOutput`. We might in the future, but right now we *only* use this field when parsing the `DenoOutput` fails. (See `OnlyTsLibVersion` below.) But I thought it'd be confusing to *not* add this field here, since it's now required in the `DenoOutput` type in typescript.

jkarni (approved):

PR #336, opened 2023-11-01
Hide non-runnable projects from `garn run --help`

### LINKED ISSUE #297: Move targets/commands to "Unavailable" section if they don't fulfill requirements.
If I run `garn [enter|run|build|check] --help`, I'll see all projects. Instead, I should only see the projects that have the appropriate type. Others should be listed in "Unavailable commands" section, along with a description of the reason (e.g., "not an executable or project with default executable").

This is related to #296 , which is about what happens when you try running the command rather than when you see the help.

alexdavid:
Resolves #297

### DIFF
--- a/src/Garn/GarnConfig.hs
+++ b/src/Garn/GarnConfig.hs
@@ -83,7 +83,7 @@ data ProjectTarget = ProjectTarget
   { description :: String,
     packages :: [String],
     checks :: [String],
-    executable :: Bool
+    runnable :: Bool
   }
   deriving (Generic, FromJSON, Eq, Show)
 
--- a/src/Garn/GarnConfig.hs
+++ b/src/Garn/GarnConfig.hs
@@ -82,7 +82,8 @@ data TargetConfig
 data ProjectTarget = ProjectTarget
   { description :: String,
     packages :: [String],
-    checks :: [String]
+    checks :: [String],
+    executable :: Bool
   }
   deriving (Generic, FromJSON, Eq, Show)
 
--- a/src/Garn/Optparse.hs
+++ b/src/Garn/Optparse.hs
@@ -81,7 +81,7 @@ buildCommand targets =
 
 runCommand :: Targets -> Parser WithGarnTsCommand
 runCommand targets =
-  Run <$> commandOptionsParser (Map.filter isExecutable targets) <*> argvParser
+  Run <$> commandOptionsParser (Map.filter isRunnable targets) <*> argvParser
   where
     argvParser :: Parser [String]
     argvParser = many $ strArgument $ metavar "...args"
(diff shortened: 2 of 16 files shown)


src/Garn/Optparse.hs
```diff
@@ -103,6 +103,11 @@ isProject = \case
   TargetConfigProject _ -> True
   TargetConfigExecutable _ -> False
 
+isExecutable :: TargetConfig -> Bool
+isExecutable = \case
+  TargetConfigExecutable _ -> True
+  TargetConfigProject p -> executable p
```
soenkehahn, line 109:
```suggestion
  TargetConfigProject projectTarget -> executable projectTarget
```

src/Garn/GarnConfig.hs
```diff
@@ -82,7 +82,8 @@ data TargetConfig
 data ProjectTarget = ProjectTarget
   { description :: String,
     packages :: [String],
-    checks :: [String]
+    checks :: [String],
+    executable :: Bool
```
soenkehahn, line 86:
```suggestion
    runnable :: Bool
```

Or maybe `isExecutable`? Or `hasDefaultExecutable`? I think `executable` is not ideal, since we do have a type called `Executable` and that might be confusing. I think I'd go with `runnable`.

(And then rename `isExecutable` in `OptParse` to `isRunnable`. I think `isExecutable` there is also kind of a confusing name.)

test/spec/RunSpec.hs
```diff
@@ -168,3 +168,36 @@ spec =
...
+        describe "top-level projects" $ do
+          it "shows runnable projects in the help" $ do
+            writeFile "garn.ts" $
+              unindent
+                [i|
+                  import * as garn from "#{repoDir}/ts/mod.ts"
+
+                  export const myProject = garn.mkProject({
+                    description: "a runnable project",
+                    defaultExecutable: garn.shell`echo runable`,
```
soenkehahn, line 181:
```suggestion
                    defaultExecutable: garn.shell`echo runnable`,
```

test/spec/RunSpec.hs
```diff
@@ -168,3 +168,36 @@ spec =
...
+                |]
+            output <- runGarn ["run", "--help"] "" repoDir Nothing
+            stdout output
+              `shouldMatch` unindent
+                [i|
+                  Available commands:
+                    myProject.*
+                |]
+
+          it "does now show non-runnable projects in the help" $ do
```
soenkehahn, line 192:
```suggestion
          it "does not show non-runnable projects in the help" $ do
```

test/spec/RunSpec.hs
```diff
@@ -168,3 +168,36 @@ spec =
...
+            writeFile "garn.ts" $
+              unindent
+                [i|
+                  import * as garn from "#{repoDir}/ts/mod.ts"
+
+                  export const myProject = garn.mkProject({
+                    description: "a runnable project",
```
soenkehahn, line 199:
```suggestion
                    description: "not runnable",
```

soenkehahn (approved):
Very nice! Modulo comments.

PR #337, opened 2023-11-01
Add smoke test for go and haskell initializers (and fix the go one)

### LINKED ISSUE #333: Add smoke test for go and haskell initializers

soenkehahn:
Fixes #333.

### DIFF
--- a/test/spec/ExampleSpec.hs
+++ b/test/spec/ExampleSpec.hs
@@ -4,10 +4,9 @@
 module ExampleSpec where
 
 import Control.Concurrent (threadDelay)
-import Control.Concurrent.Async (waitEitherCatch, withAsync)
-import Control.Exception (SomeException, bracket, catch, throwIO)
+import Control.Exception (catch)
 import Control.Lens ((^.))
-import Control.Monad (forM_, when)
+import Control.Monad (forM_)
 import Data.String.Conversions (cs)
 import Data.String.Interpolate (i)
 import Data.Text (Text)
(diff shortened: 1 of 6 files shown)


test/spec/TestUtils.hs
```diff
@@ -186,3 +188,44 @@ onTestFailureLogger test = do
...
+      when (not running) $ do
+        threadDelay 100000
+        waitUntilRunning
+
+    isRunning = do
+      Exit c <-
+        cmd
+          "curl --silent localhost:8777/base.ts"
+          (EchoStdout False)
+      pure $ c == ExitSuccess
```
soenkehahn, line 231: This just moved here unchanged from `ExampleSpec`.

ts/go/mod.ts
```diff
@@ -10,6 +10,8 @@ import { GOMOD2NIX_REPO } from "./consts.ts";
 const getGoModNixToml = (src: string): NixExpression => {
   const gen = new Deno.Command("nix", {
     args: [
+      "--extra-experimental-features",
+      "nix-command flakes",
```
soenkehahn, line 14: This is a fix for the go initializer that the test revealed.

test/spec/InitSpec.hs
```diff
@@ -4,90 +4,144 @@ module InitSpec where
 
 import Data.String.Interpolate (i)
 import Data.String.Interpolate.Util (unindent)
+import Development.Shake (StdoutTrim (..), cmd, cmd_)
```
soenkehahn, line 7: This file is probably best reviewed with white space changes switched off.

test/spec/ExampleSpec.hs
```diff
@@ -51,28 +30,6 @@ retryGet url = fmap cs <$> go 1000
...
-        threadDelay 100000
-        waitUntilRunning
-
-    isRunning = do
-      Exit c <-
-        cmd
-          "curl --silent localhost:8777/base.ts"
-          (EchoStdout False)
-      pure $ c == ExitSuccess
-
```
soenkehahn, line 75: All this just moved to `TestUtils` unchanged.

test/spec/InitSpec.hs
```diff
@@ -4,90 +4,144 @@ module InitSpec where
...
+            }
+          |]
+        output <- runGarn ["init"] "" repoDir Nothing
+        onTestFailureLog output
+        stderr output `shouldBe` "[garn] Creating a garn.ts file\n"
+        let check = do
+              output <- runGarn ["run", "myProject.start"] "" repoDir Nothing
+              onTestFailureLog output
+              stdout output `shouldContain` "starting my project..."
+        check
```
alexdavid, line 131: I think this check will cause the problem of failing anytime we introduce a breaking change to the IPC (which is what I ran into on #336). I'm wondering if we should just remove these non-replaced checks

alexdavid (approved):
This looks good, but I think we need to remove the double check now that I think about it

alexdavid (approved):

PR #341, opened 2023-11-02
Simpler versions of `addCheck` and `addExecutable`

soenkehahn:
This is a somewhat drastic change, but the new overloads feel much better than the backtick version. I *think* for now we should still support the backtick version for cases where users want to splice in packages.

Also while I *think* this is a backwards compatible change, maybe we should still bump the ts lib version? WDYT?

### DIFF
--- a/test/spec/CheckSpec.hs
+++ b/test/spec/CheckSpec.hs
@@ -44,7 +44,7 @@ spec = do
                 src: "."
               })
                 .withDevTools([garn.mkPackage(nixRaw`pkgs.hlint`)])
-                .addCheck("hlint")`hlint *.hs`;
+                .addCheck("hlint", "hlint *.hs");
             |]
           output <- runGarn ["check", "haskell"] "" repoDir Nothing
           onTestFailureLog output
@@ -81,9 +81,9 @@ spec = do
               export const failing = garn.mkProject(
                 { description: 'Failing Project' },
                 {
-                  check1: garn.check`echo ABC`,
-                  check2: garn.check`echo DEF && false`,
-                  check3: garn.check`echo GHI`,
+                  check1: garn.check("echo ABC"),
+                  check2: garn.check("echo DEF && false"),
+                  check3: garn.check("echo GHI"),
                 }
               );
             |]
@@ -107,6 +107,27 @@ spec = do
           output <- runGarn ["check", "myProject"] "" repoDir Nothing
           onTestFailureLog output
           stderr output `shouldContain` "hello world"
+          exitCode output `shouldBe` ExitFailure 1
+
+        it "allows to use backtick syntax" $ \onTestFailureLog -> do
+          writeHaskellProject repoDir
+          writeFile
+            "garn.ts"
+            [i|
+              import * as garn from "#{repoDir}/ts/mod.ts"
+              import { nixRaw } from "#{repoDir}/ts/nix.ts";
+
+              export const haskell = garn.mkProject(
+                {
+                  description: "",
+                  defaultEnvironment: garn.emptyEnvironment,
+                },
+                {},
+              ).addCheck("test-check")`echo error; false`;
+            |]
+          output <- runGarn ["check", "haskell"] "" repoDir Nothing
+          onTestFailureLog output
+          stderr output `shouldContain` "error"
           exitCode output `shouldBe` ExitFailure 1
 
         describe "exit-codes" $ do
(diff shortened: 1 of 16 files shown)


jkarni:
Can you remind me why we can’t just use normal template literals (by writing custom toString methods that generate the nix string) and get rid of the tag function altogether?

alexdavid:
> Can you remind me why we can’t just use normal template literals (by writing custom toString methods that generate the nix string) and get rid of the tag function altogether?

The issue with this is that we can't do any escaping, a shell function that includes a quote, backslash, or `${...}` would be indistinguishable from an actual nix expression splatted in.

test/spec/CheckSpec.hs
```diff
@@ -109,6 +109,27 @@ spec = do
...
+              import * as garn from "#{repoDir}/ts/mod.ts"
+              import { nixRaw } from "#{repoDir}/ts/nix.ts";
+
+              export const haskell = garn.mkProject(
+                {
+                  description: "",
+                  defaultEnvironment: garn.emptyEnvironment,
+                },
+                {},
+              ).addCheck("test-check")`echo error; false`;
```
alexdavid, line 126: Maybe this test should actually splat in a package to help document why we want to support this syntax still?

ts/project.ts
```diff
@@ -163,58 +179,80 @@ const proxyEnvironmentHelpers = (): ProjectHelpers => ({
...
       };
-    };
-    markAsMayNotExport(templateLiteralFn, (exportName: string) =>
-      [
-        `${exportName} exports the return type of "addExecutable", but this is not the proper usage of addExecutable.`,
-        'Did you forget the template literal? Example usage: project.addExecutable("executable-name")`shell script to run`',
-      ].join(" "),
-    );
-    return templateLiteralFn;
+    } else {
```
alexdavid, line 198: Nit, but maybe we remove this else? And the one in `addCheck` below?

ts/project.ts
```diff
@@ -163,58 +179,80 @@ const proxyEnvironmentHelpers = (): ProjectHelpers => ({
...
-      ...args: Array<string>
-    ) => {
-      const newExecutable = { [name]: this.shell(s, ...args) } as {
-        [n in Name]: Executable;
-      };
+    if (script != null) {
+      const newExecutable = { [name]: this.shell(script) };
       return {
         ...this,
         ...newExecutable,
```
alexdavid, line 196: Interesting that you didn't run into the same `name` becoming `string` type degradation that i ran into with the extra function layer. Since this is the case, should we just inline it here?

```suggestion
        [name]: this.shell(script),
```

alexdavid (approved):
This looks great.

I was less concerned about `shell` and `check` using the template literals (in fact I think they're really  nice there), than the `addCheck` and `addExecutable` since it's kind of strange after a function call. But I don't feel strongly.

I think this PR should update the code on the website, examples, and readme before merging.

jkarni:
> The issue with this is that we can't do any escaping. A shell snippet that includes a quote, backslash, or ${...} would be indistinguishable from an actual nix expression splatted in.

Why though? We could have the toString be whatever we want - $$$THIS-IS-PATH-somepath$$$ - and then process the whole string before giving it to nix.

alexdavid:
Yeah, that's true, but it seems a little sad that we don't keep it structural in my opinion, especially since the user provides the structural data and we intentionally replace it with some magic string. Since we realized that splatting in packages is less of a common flow I think  having a specific tag template literal function that we can use feels like the right solution to me, but I don't feel too strongly if you want to keep them strings:

```typescript
something.addExecutable("run-cowsay", nix`${pkgs.cowsay}/bin/cowsay moo`)
```

ts/project.ts
soenkehahn: I think this might be because we updated the `deno` version in between?

Inlined in 4b19957c.

soenkehahn:
> I think this PR should update the code on the website, examples, and readme before merging.

Good call, I updated the website and the readme in e673f18 (this one deserves scrutiny, since we don't typecheck this automatically anywhere) and updated the examples in 0459bef.

ts/project.ts
soenkehahn: Done in fd157e61.

test/spec/CheckSpec.hs
soenkehahn: Yeah, good call! I did that in a0f09c5f. And also had to actually make it work. :)

alexdavid (approved):
Pushed a nit review commit. Looks great!

PR #349, opened 2023-11-03
Better handling of user interrupts (fixes #299)

### LINKED ISSUE #299: Better handling of user interrupts.
See the following interaction, where I start a server, and then Ctrl-C it:

```
% garn run server/dev                                                                 
warning: Git tree '/home/jkarni/garnix/garn' is dirty
Backend listening on port 3000
^Csignal: interrupt
garn: callProcess: nix "run" "--extra-experimental-features" "flakes nix-command" "--print-build-logs" ".#server/dev" "--" (exit 1): failed
```

In my opinion, two things are wrong here: we should *never* surface such an ugly error message, even if the nix run command failed by itself, and in the particular case of Ctrl-C, we shouldn't surface anything at all, besides what the underlying program does.

sol:


cabal.project
```diff
@@ -1 +1,2 @@
 packages: garn.cabal
+tests: True
```
sol, line 2: Does this affect production builds? If yes, then you may not want this change.

src/Garn.hs
```diff
@@ -48,7 +48,7 @@ runWith env (WithGarnTsOpts garnConfig opts) = do
   case opts of
     Gen -> pure ()
     Run (CommandOptions {..}) argv -> do
-      callProcess "nix" $ ["run"] <> nixArgs <> [".#" <> asNixFacing target, "--"] <> argv
+      rawSystem "nix" (["run"] <> nixArgs <> [".#" <> asNixFacing target, "--"] <> argv) >>= throwIO
```
soenkehahn, line 51:
```suggestion
      exitCode <- rawSystem "nix" (["run"] <> nixArgs <> [".#" <> asNixFacing target, "--"] <> argv)
      exitWith exitCode
```

I think this makes it a bit clearer that this is not an exceptional case where we throw, but instead just exiting with the proper exit code.

(`rawSystem` seems fine to use.)

.gitattributes
```diff
@@ -0,0 +1,2 @@
+*.cabal linguist-generated
+/examples/init/foo.cabal -linguist-generated
```
soenkehahn, line 2: This is a hand-written file.
```suggestion
```

cabal.project
soenkehahn: Our production builds are done with `nix` and this doesn't seem to affect that. What's the workflow you're using where this is beneficial?

soenkehahn (approved):
Nice, thanks for the PR! Left some comments.

src/Garn.hs
sol: This certainly works for me.

.gitattributes
sol: To be clear, what I am saying here is that all `.cabal` files are generated, except for `/examples/init/foo.cabal` (note the minus sign in `-linguist-generated`).

Does this make sense?

cabal.project
sol: This ensures, that a `cabal build` always builds all test suites, without having to do a `cabal configure --enable-tests` first.  It also protects you against https://github.com/haskell/cabal/issues/5079, https://github.com/haskell/cabal/issues/7883 (yes, that's true, `cabal test all` does not necessarily build all your tests, https://github.com/haskell/cabal/issues/5079#issuecomment-751373847).

With nix, I am not sure how essential this still is for my workflow. So far I used a combination of `cabal`-targets, `just`-targets and `nix profile install` (which still fails for me, but populates the nix store, which I then can reference manually).

Not sure what exactly `nix profile install` does, but given that the `justfile` runs `cabal`-targets, I would probably tend to `tests: true` so that `cabal` always solves for a consistent set of dependencies, not sure.

sol:
@soenkehahn thanks for the review. I am away from the computer right now, I will do a rebase later.

sol:
>I will do a rebase later.

Done.

cabal.project
soenkehahn: Sounds good. When I'm working on `garn`, `nix` provides all dependencies, so hopefully the cabal solver just finds all the necessary dependencies already installed and just uses them. But being explicit about that we *always* want it to solve for the test-suite seems good.

.gitattributes
soenkehahn: Ah, thanks for explaining! Looks good to me then.

soenkehahn:
@sol: I think the PR makes sense, so I'll merge it. I'm a bit confused by the title: before `ctrl+c` was also delegated to the child process. Did you experience a case where it wasn't?

sol:
> I'm a bit confused by the title: before `ctrl+c` was also delegated to the child process. Did you experience a case where it wasn't?

@soenkehahn see #299.

Also note that before this PR, any process that would call `garn run` could not detect when the subprocess was canceled due to ctrl-c, which can be a problem in some situations.

PR #350, opened 2023-11-03
Improve copy button

### LINKED ISSUE #346: It should be easy to copy & paste the install command on Linux
On https://garn.io/docs/getting_started/, when I tripple-click the install command, then it also selects that "copy" link, and pasting will give me:
```
sh <(curl --proto '=https' --tlsv1.2 -sSf https://garn.io/install.sh)Copy
```

alexdavid:
* Prevents user select of copy button
* Prevents wrapping of copy button outside of box on narrow screens
* Uses <button> instead of <span> for copy button
* Restores "copy" text after a few seconds when clicking copy

Resolves #346

### DIFF
--- a/website/src/components/BashSnippetToCopy.tsx
+++ b/website/src/components/BashSnippetToCopy.tsx
@@ -1,13 +1,19 @@
-import { useState } from "react";
+import { useState, useEffect } from "react";
 import { CopyToClipboard } from "react-copy-to-clipboard";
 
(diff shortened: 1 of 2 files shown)


soenkehahn:
>     * Uses  instead of  for copy button

I think this is missing a few words?

soenkehahn (approved):

alexdavid:
> I think this is missing a few words?

Looks like it's github's automatic description inference from my commit message is broken. The commit has the correct text

PR #352, opened 2023-11-03
Lock down installed version of garn

alexdavid:


soenkehahn (approved):
Yeah, I think this makes sense, we should merge. And also deploy the website. (Maybe it'd be nice to test the installer manually after doing that. So maybe do that on Monday?)

PR #354, opened 2023-11-03
Keep `NixExpression`s as AST until serialization

alexdavid:
Given our talk today about implementing flake input helpers in `nix.ts` I felt like this was a good refactor to do now. By keeping everything in an AST we can start to walk the tree and more accurately understand scopes (and find duplication in the future).

### DIFF
--- a/ts/nix.ts
+++ b/ts/nix.ts
@@ -34,11 +34,21 @@ export type NixStrLitInterpolatable =
  *
  * It is not advised to construct this type but instead use `nixRaw` or `nixStrLit`.
  */
-export type NixExpression =
+export type NixExpression = { [__nixExpressionTag]: null } & (
   | { type: "raw"; raw: InterpolatedString<NixExpression> }
   | { type: "list"; elements: Array<NixExpression> }
   | { type: "attrSet"; elements: Record<string, NixExpression> }
-  | { type: "strLit"; str: InterpolatedString<NixExpression> };
+  | { type: "strLit"; str: InterpolatedString<NixExpression> }
+);
+
+/**
+ * Tag for `NixExpression`.
+ *
+ * This symbol is deliberately not exported, since you're not supposed to create
+ * `NixExpression`s outside of this module. Instead use helper functions from
+ * here.
+ */
+const __nixExpressionTag = Symbol("NixExpression is opaque!");
 
 /**
  * A template literal function to construct `NixExpression`s from raw strings.
@@ -74,6 +84,7 @@ export function nixRaw(
   ...interpolations: Array<NixExpression>
 ): NixExpression {
   return {
+    [__nixExpressionTag]: null,
     type: "raw",
     raw:
       typeof s === "string"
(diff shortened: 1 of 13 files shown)


ts/nix.ts
```diff
@@ -132,37 +151,148 @@ export function nixStrLit(
...
+  return astToHumanReadable(nixExpr.__ast);
+}
+
+/**
+ * Converts a `NixExpression` to a string representing that Nix expression.
+ *
+ * It is advised to defer calling this unless actually serializing to disk.
+ * This way Nix expressions remain as the type `NixExpression`.
+ */
+export function toNixString(nixExpr: NixExpression): string {
```
soenkehahn, line 199:
```suggestion
export function renderNixExpression(nixExpr: NixExpression): string {
```

I think `toNixString` is confusing, since it could talk about nix string literals.

ts/nix.ts
```diff
@@ -132,37 +151,148 @@ export function nixStrLit(
...
+              acc + "${" + astToNixString(astNode) + "}" + escape(str),
+            escape(node.str.initial),
+          ) +
+          '"'
+        );
       }
-    }, "") +
-    '"';
-  return { rawNixExpressionString };
+    }
```
soenkehahn, line 234: Can we use `utils.checkExhaustiveness` here to make sure we cover all ast node types?

ts/nix.ts
```diff
@@ -132,37 +151,159 @@ export function nixStrLit(
...
+ *   // i is [1, 2]
+ *   // but nothing at the type level enforces that s.length - 1 == i.length
+ *
+ *   // Using `interpolatedStringFromTemplate` we can convert to a safer `InterpolatedString`:
+ *   const interpolatedString = interpolatedStringFromTemplate(s, i);
+ *   // interpolatedString is { initial: "foo", rest: [[1, "bar"], [2, "baz"]] }
+ * }
+ * ```
+ */
+type InterpolatedString<T> = {
```
soenkehahn, line 262: Should this type and its helper functions go into a separate module?

soenkehahn (commented):
This looks great! I left some comments and would love to take another look once those are addressed.

ts/nix.ts
```diff
@@ -64,13 +70,18 @@ export function nixRaw(
...
-  return { rawNixExpressionString };
+  if (typeof s === "string") {
+    return { __ast: { type: "raw", raw: interpolatedStringFromString(s) } };
+  }
+  return {
+    __ast: {
+      type: "raw",
+      raw: interpolatedStringFromTemplate(
+        s,
+        interpolations.map((i) => i.__ast),
```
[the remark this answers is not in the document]
alexdavid: Yeah I wasn't sure if I liked the extra wrapper or not. I'll remove it, we can mention that you're on your own if you start messing with the AST

ts/nix.ts
alexdavid: Yeah, although the return type of this function does enforce this, `checkExhaustiveness` probably gives a better type error location though

soenkehahn: Oh, I see. Never mind then!

soenkehahn (approved):
Nice! I added a tag to `NixExpression` in a review commit. That doesn't *guarantee* that no-one outside of this module creates a `NixExpression`, but it does make it a little bit harder. (Except for splatting, I guess. :shrug:) TIOLI!

PR #360, opened 2023-11-06
Move flake deps into NixExpression AST

alexdavid:


ts/nix.ts
```diff
@@ -163,6 +170,21 @@ export function nixStrLit(
   };
 }
 
+/**
+ * Returns a `NixExpression` which represents a dependency of the file file
```
jkarni, line 174:
```suggestion
 * Returns a `NixExpression` which represents a dependency of the file 
```

Though I don't understand this...

ts/nix.ts
```diff
@@ -163,6 +170,21 @@ export function nixStrLit(
...
+ * this expression is embedded in.
+ */
+export function nixFlakeDep(name: string, dep: FlakeDep): NixExpression {
```
soenkehahn, line 177:
```suggestion
/**
 * Returns a `NixExpression` that renders as an identifier that refers to a flake input. At the same time it registers the flake input as a dependency, so that it'll be included in the inputs of the generated flake file. See `renderFlakeFile` for an example.
 */
export function nixFlakeDep(name: string, dep: FlakeDep): NixExpression {
```

ts/nix.ts
```diff
@@ -174,6 +196,72 @@ export function toHumanReadable(nixExpr: NixExpression): string {
   return renderInterpolatedString(nixExpr.str, toHumanReadable);
 }
 
+function collectFlakeDeps(nixExpr: NixExpression): Record<string, FlakeDep> {
+  const collect = (arr: Array<NixExpression>) =>
+    arr.reduce(
+      (acc, nixExpr) => ({ ...acc, ...collectFlakeDeps(nixExpr) }),
```
soenkehahn, line 202: Should we throw here if there's duplicate keys that don't map to the same exact flake url? That seems like a situation where we want to abort.

ts/nix.ts
alexdavid: Whoops, this was supposed to say `flake file`. The idea is that these flake dependencies are captured sort of like a side-effect in the AST to be pulled out at the end when generating the flake file. This solves a few problems, one of being able to declare flake file dependencies in your `garn.ts` (although this currently doesn't have a nice API yet), but the other is to remove the need to install gomod2nix and npmlock2nix for every garn project. Now they are only pulled in as needed

ts/nix.ts
```diff
@@ -174,6 +196,72 @@ export function toHumanReadable(nixExpr: NixExpression): string {
...
+          : `inputs.${name} = { url = ${quoteStr(dep.url)}; flake = false; };`,
+      )
+      .join("\n") +
+    "\noutputs = { " +
+    ["self", ...Object.keys(flakeDeps)].join(", ") +
+    " }:" +
+    renderNixExpression(nixExpr) +
+    ";\n}"
+  );
+}
```
soenkehahn, line 263: Could we add some tests for this? Probably in typescript, I think. Some simple test (like the one from the docstring here), but also some tests where we call `nixFlakeDep` in a deeper nested structure, like in a list or an attribute set or both.

ts/javascript/mod.ts
```diff
@@ -52,9 +58,13 @@ export function mkNpmProject(args: {
   node_modules: Package;
 } {
   const { pkgs, nodejs } = fromNodeVersion(args.nodeVersion);
+  const npmlock2nixRepo = nixFlakeDep("npmlock2nix-repo", {
+    url: "github:nix-community/npmlock2nix?rev=9197bbf397d76059a76310523d45df10d2e4ca81",
+    flake: false,
+  });
```
soenkehahn, line 64: When ignoring the internals of the `nix.ts` module, the API of `nixFlakeDep` is a bit weird. Why do I pass in the name as a parameter and then `url` and `flake` as an object? Maybe instead have it be this:

```suggestion
  const npmlock2nixRepo = nixFlakeDep(
    "npmlock2nix-repo",
    "github:nix-community/npmlock2nix?rev=9197bbf397d76059a76310523d45df10d2e4ca81",
    flake: false,
  );
```

I.e.: `nixFlakeDep` should have type `(name: string, url: string, flake: boolean = false): NixExpression`.

soenkehahn (approved):
Overall this seems really nice. Left some comments.

I'm a bit concerned since this is a pattern that I haven't seen in other code-bases/AST definitions. I *do* think that most ASTs are probably mostly for parsing and not for codegen. So maybe that's why? In any case, this concern is too vague to address. :shrug:

ts/nix.ts
alexdavid: Yeah this is probably best for now. I was assuming we'd eventually throw in the functions that build up the AST (as we transition away from nixRaw).

ts/javascript/mod.ts
alexdavid: Are there other potential options that go on a flake input other than `url` and `flake`? I kind of assumed there might be others which is why I wanted the type, but I couldn't find documentation on this. If not maybe your proposal is best, but also the `false` would not be annotated as it is in your example which makes it less readable. But maybe that's ok?

alexdavid: Thinking about this more, while it is nice to re-use the type (that I would hope eventually includes all the valid values of a flake input), the fact that some of these keys are optional are a bit annoying (see `flakeDepEq` and it's equality check of `flake` parameter). Maybe we should have separate types for the input that is more lax than the type in the AST? Would like to hear your thoughts.

soenkehahn: (Talked about this offline.) I think it's fair that the `nix` module exposes an API that is very similar to what `nix` flakes look like. So I think the current version is best for now.

PR #375, opened 2023-11-07
Enable github CI for all (including external) PRs

soenkehahn:
Currently, our github CI doesn't run for PRs from forked repos, e.g. for this one:
https://github.com/garnix-io/garn/pull/349

This hopefully fixes that.

I think we still have to manually approve CI runs from external forks. But that seems sensible for now.

Update: I first tried out `on: [push, pull_request]`, but that means that for internal PRs we always get duplicated runs, since they're both a PR *and* a push to a local branch. I changed the config to now:

- Enable CI only for pushes to the `main` branch, *and*
- enable CI for all PRs.

But this means that we won't get CI runs for pushes to branches other then `main` in this repo, if there's no PR open for that branch. So if you want a CI run, you'd *have* to open a PR (possibly a draft PR). I think this is acceptable. WDYT?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,6 +1,8 @@
 name: CI
 
-on: push
+on:
+  - push
+  - pull_request
 
 jobs:
   catchall:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,8 +1,10 @@
 name: CI
 
 on:
-  - push
-  - pull_request
+  pull_request:
+  push:
+    branches:
+      - main
 
 jobs:
   catchall:


alexdavid (approved):

sol:
> Currently, our github CI doesn't run for PRs from forked repos, e.g. for this one: #349
> 
> This hopefully fixes that.
> 
> I think we still have to manually approve CI runs from external forks. But that seems sensible for now.
> 
> Update: I first tried out `on: [push, pull_request]`, but that means that for internal PRs we always get duplicated runs, since they're both a PR _and_ a push to a local branch. I changed the config to now:
> 
> * Enable CI only for pushes to the `main` branch, _and_
> * enable CI for all PRs.
> 
> But this means that we won't get CI runs for pushes to branches other then `main` in this repo, if there's no PR open for that branch. So if you want a CI run, you'd _have_ to open a PR (possibly a draft PR). I think this is acceptable. WDYT?


That's basically what I have been doing for most of my repositories for quite a while now, and it works very well for me.

https://github.com/sol/hpack/blob/main/.github/workflows/build.yml#L3-L15

You may also want that first three lines, so that in-progress builds are canceled when you add additional commits to a PR.

PR #376, opened 2023-11-07
Implement `addPackage` helper

### LINKED ISSUE #369: Support specifying manual build commands via an `addPackage(name, commands)`

alexdavid:
Supports specifying named manual build commands on a project.

Resolves #369

### DIFF
--- a/test/spec/BuildSpec.hs
+++ b/test/spec/BuildSpec.hs
@@ -107,6 +107,27 @@ spec = do
         stderr output `shouldMatch` "project.short \\s+ Builds short command"
         stderr output `shouldMatch` "project.longer \\s+ Builds # this is some longe\\.\\.\\."
 
+      describe "addPackage" $ do
+        it "builds packages within the project's default environment" $ \runGarn -> do
+          writeFile
+            "garn.ts"
+            [i|
+              import * as garn from "#{repoDir}/ts/mod.ts"
+
+              const env = garn.mkEnvironment(undefined, garn.nix.nixStrLit`
+                mkdir dist
+                echo build-content > dist/build-artifact
+              `);
+              export const project = garn.mkProject({
+                description: "",
+                defaultEnvironment: env,
+              }, {})
+                .addPackage("build", "mv dist/build-artifact $out/build-artifact");
+            |]
+          output <- runGarn ["build", "project"]
+          readFile "result/build-artifact" `shouldReturn` "build-content\n"
+          exitCode output `shouldBe` ExitSuccess
+
       describe ".build" $ do
         it "builds manually specified packages" $ \runGarn -> do
           writeFile
--- a/ts/project.ts
+++ b/ts/project.ts
@@ -1,7 +1,7 @@
 import "./internal/registerInternalLib.ts";
 
 import { Check, mkCheck } from "./check.ts";
-import { Environment } from "./environment.ts";
+import { Environment, emptyEnvironment } from "./environment.ts";
 import { Executable, mkShellExecutable } from "./executable.ts";
 import { hasTag } from "./internal/utils.ts";
 import { NixStrLitInterpolatable } from "./nix.ts";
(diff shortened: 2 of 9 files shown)


test/spec/BuildSpec.hs
```diff
@@ -42,6 +42,42 @@ spec = do
...
+                  defaultEnvironment: env,
+                }, {})
+                  .addPackage("build", "mv dist/build-artifact $out/build-artifact");
+              |]
+            output <- runGarn ["build", "project"] "" repoDir Nothing
+            onTestFailureLog output
+            readFile "result/build-artifact" `shouldReturn` "build-content\n"
+            exitCode output `shouldBe` ExitSuccess
+
+          it "allows adding packages to projects without a default environment" $ \onTestFailureLog -> do
```
soenkehahn, line 67:
```suggestion
          it "allows adding packages to projects" $ \onTestFailureLog -> do
```

And make this the first test case of the two added ones? This feels like the more basic test.

test/spec/BuildSpec.hs
```diff
@@ -42,6 +42,42 @@ spec = do
           stderr output `shouldContain` "Couldn't match type"
           exitCode output `shouldBe` ExitFailure 1
 
+        describe "addPackage" $ do
+          it "allows adding packages to projects within the project's environment" $ \onTestFailureLog -> do
```
soenkehahn, line 46:
```suggestion
          it "builds packages within the project's default environment, if it exists" $ \onTestFailureLog -> do
```

ts/project.ts
```diff
@@ -247,6 +268,20 @@ const proxyEnvironmentHelpers = (): ProjectHelpers => ({
...
   },
 
+  addPackage<T extends Project, Name extends string>(
+    this: T,
+    name: Name,
+    pkg: string,
+  ) {
+    const build = this.defaultEnvironment
+      ? this.build(pkg)
+      : emptyEnvironment.build(pkg);
```
soenkehahn, line 278: For `Project.addCheck` and `Project.addExecutable` we decided that -- if there's no default environment -- we crash. (I wonder whether we should reconsider that.) But why is `Project.addPackage` different?

ts/project.ts
```diff
@@ -247,6 +268,20 @@ const proxyEnvironmentHelpers = (): ProjectHelpers => ({
...
+    return {
+      ...this,
+      ...({ [name]: build } as { [n in Name]: Package }),
+    };
+  },
+
```
soenkehahn, line 284: You didn't implement the backtick syntax that we still support for `Project.addCheck` & `Project.addExecutable`. I'm assuming that's not an oversight, but just something that we don't want to support for `Project.addPackage` right now, because in general we want a better API for that. Is that correct or is that just an oversight?

soenkehahn (commented):
Overall looks good to me. Left some comments and questions though.

ts/project.ts
alexdavid: Yeah that's kind of what I was thinking, but don't feel strongly. I was assuming the template literal syntax was being deprecated and therefore we shouldn't add to new methods, but let me know if you think we should be consistent until we fully remove it

ts/project.ts
alexdavid: Ah my bad, for some reason I thought we only had that restriction on `.build` and `.shell` but not on the `.add*` methods. Will update

ts/project.ts
soenkehahn: No, I think that's a good call!

soenkehahn (approved):
Approving now, modulo the outstanding unresolved conversations.

soenkehahn:
Oh, now that we have a changelog file, there should probably be a note added to that in this PR.

test/spec/BuildSpec.hs
alexdavid: This test case was deleted since we don't support `addPackage` on non-default environments

Issue #339, opened 2023-11-02
Export `pkgs.nodePackages` (for e.g. `prettier`) in `nixpkgs.ts`


soenkehahn:

jkarni:
How will this work? Will it be exported as a nested record? (It feels like we can’t export all these packages as the top level without a big risk of conflict.)

alexdavid:
The implementation I have in #382 breaks each sub package up into its own file and uses `export * from "..."`. This way `pkgs` matches as close as possible to the real nixpkgs in terms of structure, and we don't have conflicts.

PR #382, opened 2023-11-09
Implement nested packages

alexdavid:
Implements #339 

Still need to figure out a good way of testing this. I think maybe we should have a feature test that tries to import from the file server and uses a node package for example?


ts/nixpkgs.ts
```diff
@@ -0,0 +1 @@
+export * from "./internal/nixpkgs/mod.ts";
```
alexdavid, line 1: This will cause type errors locally until running codegen. Should we ignore this file again, or check in the codegen files?

soenkehahn: Ignoring is not really an option, since we need this file in the repo and it's hand-written, right?

We could generate this file too. Just `writeFile "ts/nixpkgs.ts" "export * from ...\n"`.

Is there a real downside to checking in the generated files though? Maybe we need a CI check that they're in sync then. But I wouldn't be against that. (But we could defer that to another PR.)

test/spec/Garn/CodeGenSpec.hs
```diff
@@ -2,161 +2,372 @@
...
+
+              export const a_package = mkPackage(
+                nixRaw`pkgs.a_package`,
+                "",
+              );
+            |]
+
+testScanPackages :: String -> String -> String -> Map String PkgInfo -> SpecWith ()
+testScanPackages description pkgs pkgSpec expectedOutput =
+  it description $ do
```
soenkehahn, line 359:
```suggestion
testScanPackages :: String -> String -> Map String PkgInfo -> SpecWith ()
testScanPackages pkgs pkgSpec expectedOutput =
```

And then just call it with `it "..." $ testScanPackages ...` instead?

soenkehahn: Same for `testWritePkgFiles`.

test/spec/Garn/CodeGenSpec.hs
```diff
@@ -2,161 +2,372 @@
...
+    withModifiedEnvironment [("NIX_CONFIG", "experimental-features =")] $ do
+      system <- currentSystem
+      actualOutput <- scanPackages system pkgs pkgSpec
+      actualOutput `shouldBe` expectedOutput
+
+testWritePkgFiles :: String -> Map String PkgInfo -> ((FilePath -> IO String) -> IO ()) -> SpecWith ()
+testWritePkgFiles description pkgInfo runTest =
+  it description $ do
+    withSystemTempDirectory "writePkgFiles-test" $ \tempDir -> do
+      writePkgFiles tempDir "path/to/garn/root" pkgInfo
```
soenkehahn, line 369:
```suggestion
      writePkgFiles tempDir "test-garn-import-root" pkgInfo
```

I was a bit confused by this path, so maybe this fake path is better?

src/Garn/CodeGen.hs
```diff
@@ -19,69 +24,111 @@ import Data.String.Interpolate (i)
...
     nixExpr =
       [i|
-        lib :
-        let mk = name: value: {
-              attribute = name;
-              description = if value ? meta.description
-                then value.meta.description
-                else null;
+        lib:
+          let
```
soenkehahn, line 94:
```suggestion
          let
```

test/spec/Garn/CodeGenSpec.hs
```diff
@@ -2,161 +2,372 @@
...
+testScanPackages :: String -> String -> String -> Map String PkgInfo -> SpecWith ()
+testScanPackages description pkgs pkgSpec expectedOutput =
+  it description $ do
+    withModifiedEnvironment [("NIX_CONFIG", "experimental-features =")] $ do
+      system <- currentSystem
+      actualOutput <- scanPackages system pkgs pkgSpec
+      actualOutput `shouldBe` expectedOutput
+
+testWritePkgFiles :: String -> Map String PkgInfo -> ((FilePath -> IO String) -> IO ()) -> SpecWith ()
+testWritePkgFiles description pkgInfo runTest =
```
soenkehahn, line 366: Before we called `deno check` to typecheck the files generated in these tests. Did you remove that intentionally? Should we get that back for the tests that use `testWritePkgFiles`?

src/Garn/CodeGen.hs
```diff
@@ -19,69 +24,111 @@ import Data.String.Interpolate (i)
 import Data.String.Interpolate.Util (unindent)
 import GHC.Generics (Generic)
 import Garn.Common (currentSystem, nixpkgsInput)
+import System.Directory (createDirectoryIfMissing, removeDirectoryRecursive)
 import WithCli (withCli)
 
+-- A specification for what packages to include in the final collection
+-- collectionSpec represents a colleciton with all derivations included by
```
soenkehahn, line 31:
```suggestion
-- collectionSpec represents a collection with all derivations included by
```

soenkehahn, line 31:
```suggestion
-- collectionSpec represents a colleciton with all top-level derivations included by
```

src/Garn/CodeGen.hs
```diff
@@ -19,69 +24,111 @@ import Data.String.Interpolate (i)
...
+-- default, but everything else excluded. This can be overridden with the
+-- attribute set.
```
soenkehahn, line 33:
```suggestion
-- given attribute set.
```

src/Garn/CodeGen.hs
```diff
@@ -19,69 +24,111 @@ import Data.String.Interpolate (i)
 import Data.String.Interpolate.Util (unindent)
 import GHC.Generics (Generic)
 import Garn.Common (currentSystem, nixpkgsInput)
+import System.Directory (createDirectoryIfMissing, removeDirectoryRecursive)
 import WithCli (withCli)
 
+-- A specification for what packages to include in the final collection
```
soenkehahn, line 30:
```suggestion
-- A nix expression that specifies which packages to include in the final collection
```

src/Garn/CodeGen.hs
```diff
@@ -19,69 +24,111 @@ import Data.String.Interpolate (i)
...
-    pure $ either error id $ eitherDecode (cs json)
+  let outDir = "ts/internal/nixpkgs"
+  removeDirectoryRecursive outDir `catch` \(_ :: IOException) -> pure ()
+  pkgs <- scanPackages system (pkgs system) pkgSpec
+  writePkgFiles outDir "../.." pkgs
+
+writePkgFiles :: String -> String -> Map String PkgInfo -> IO ()
+writePkgFiles modulePath garnLibRoot pkgs = do
+  createDirectoryIfMissing True modulePath
+  let sanitizedPkgs = Map.mapKeys sanitize pkgs
```
soenkehahn, line 69:
```suggestion
writePkgFiles modulePath garnLibRoot (Map.mapKeys sanitize -> pkgs) = do
  createDirectoryIfMissing True modulePath
```

With https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/view_patterns.html. That way we don't have the unsanitized `pkgs` in scope.

soenkehahn (commented):
Left a bunch of comments, would love to take another look later!

test/spec/Garn/CodeGenSpec.hs
alexdavid: This wasn't entirely intentional, but I assumed on CI we were typechecking the codegen files. It looks like we do call `test-ts` after `codegen` on GitHub, but at the moment this doesn't typecheck. I think maybe we should just bring over the change on #409 of adding `--check` to `test-ts` recipe. What do you think? Is there an advantage to also running typecheck in the Haskell tests?

test/spec/Garn/CodeGenSpec.hs
```diff
@@ -5,249 +5,242 @@ module Garn.CodeGenSpec (spec) where
...
-inParallel :: Spec -> Spec
-inParallel spec = do
-  runIO $ do
-    n <- getNumProcessors
-    setNumCapabilities (max 1 (n - 1))
-  parallel spec
-
 spec :: Spec
-spec = inParallel $ do
+spec = parallel $ do
```
alexdavid, line 16: As discussed, I reverted this parallel change because it wasn't working on GitHub actions.

test/spec/Garn/CodeGenSpec.hs
```diff
@@ -2,161 +2,372 @@
...
+              export const a_package = mkPackage(
+                nixRaw`pkgs.a_package`,
+                "",
+              );
+            |]
+
+testScanPackages :: String -> String -> String -> Map String PkgInfo -> SpecWith ()
+testScanPackages description pkgs pkgSpec expectedOutput =
+  it description $ do
+    withModifiedEnvironment [("NIX_CONFIG", "experimental-features =")] $ do
```
soenkehahn: This could also go into an [aroundAll_](https://www.stackage.org/haddock/lts-21.20/hspec-2.10.10/Test-Hspec.html#v:aroundAll_). That would mean it gets executed once for all the tests in this `Spec`.

soenkehahn: Very much TIOLI! It's just that you mentioned you don't like this here.

ts/nixpkgs.ts
soenkehahn: Yeah, I think that makes sense.

test/spec/Garn/CodeGenSpec.hs
soenkehahn: > [...] What do you think? Is there an advantage to also running typecheck in the Haskell tests?

I think the advantage was mostly during development, that it was nice to TDD on smaller examples and have `deno` yell at you when you weren't generating valid typescript. So I thought it was nice to have. But you were also the last person to work on this code, and if you didn't miss it, maybe it's not that useful? Up to you!

And yes, please add `--check` to `test-ts`! That makes sense in any case.

soenkehahn (approved):
This looks good to me now. There's some remaining conversations, but for all of them feel free to make a decision and merge!

test/spec/Garn/CodeGenSpec.hs
soenkehahn: > And yes, please add `--check` to `test-ts`! That makes sense in any case.

Never mind, `typescript-check` is a post-dependency of the `codegen` recipe: https://github.com/garnix-io/garn/blob/f80b691b422d7cc84a57bb2693ae9c25071c44b7/justfile#L158. So we do typecheck on CI and everytime we run `just codegen`.

PR #389, opened 2023-11-09
Better error message on IPC version mismatches

soenkehahn:


test/spec/GarnSpec.hs
```diff
@@ -123,7 +123,16 @@ spec = do
...
-                    (Internal details: Error in $: key \"tag\" not found)
+                      garn cli tool version: v0.0.15
+                      garn typescript library version: <testTsLibVersion>
+
+                      Either:
+
+                        Install version <testTsLibVersion> of the garn cli tool.
+                        See https://garn.io/docs/getting_started#updating-garn for how to update.
+
+                        Or: Use version v0.0.15 of the typescript library.
```
jkarni, line 134: A bit weird formatting. Maybe move the `Or` to a separate line like the `Either`, or have it be a list? E.g.:

```
Either:

1) Install version <testTsLibVersion> of the garn cli tool.
    (See https://garn.io/docs/getting_started#updating-garn for how to update.) Or;
2) Use version v0.0.15 of the typescript library.
```

test/spec/GarnSpec.hs
```diff
@@ -123,7 +123,16 @@ spec = do
...
+                        E.g. like this: import * as garn from "https://garn.io/ts/v0.0.15/mod.ts";
```
jkarni, line 135:
```suggestion
                        E.g.: import * as garn from "https://garn.io/ts/v0.0.15/mod.ts";
```

jkarni (approved):
Looks good!
At some point maybe it'll make sense to switch to `Doc` for the error type.

test/spec/GarnSpec.hs
soenkehahn: Done in 659bf17e.

PR #391, opened 2023-11-09
Implement running individual packages from a project

### LINKED ISSUE #371: Support specifying a single package to build on a project with `garn build project.package`

soenkehahn:
Fixes #371.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## v0.0.16
+
+- Allow to build packages that are nested within projects with `garn build projectName.packageName`.
+- Allow to build top-level packages with `garn build packageName.
+
 ## v0.0.15
 
 - Added a `--version` flag
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## v0.0.16
+## v0.0.16 (unreleased)
 
 - Allow to build packages that are nested within projects with `garn build projectName.packageName`.
 - Allow to build top-level packages with `garn build packageName.
--- a/src/Garn/GarnConfig.hs
+++ b/src/Garn/GarnConfig.hs
@@ -78,6 +78,7 @@ type Targets = Map TargetName TargetConfig
 
 data TargetConfig
   = TargetConfigProject ProjectTarget
+  | TargetConfigPackage PackageTarget
   | TargetConfigExecutable ExecutableTarget
   deriving (Generic, Eq, Show)
 
@@ -86,6 +87,11 @@ data ProjectTarget = ProjectTarget
     packages :: [String],
     checks :: [String],
     runnable :: Bool
+  }
+  deriving (Generic, FromJSON, Eq, Show)
+
+data PackageTarget = PackageTarget
+  { description :: String
   }
   deriving (Generic, FromJSON, Eq, Show)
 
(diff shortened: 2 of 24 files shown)


src/Garn/GarnConfig.hs
```diff
@@ -78,6 +78,7 @@ type Targets = Map TargetName TargetConfig
 
 data TargetConfig
   = TargetConfigProject ProjectTarget
+  | TargetConfigPackage PackageTarget
```
soenkehahn, line 81: This is an IPC-breaking change.

test/spec/BuildSpec.hs
```diff
@@ -3,6 +3,7 @@
 module BuildSpec where
 
 import Data.String.Interpolate (i)
+import Data.String.Interpolate.Util (unindent)
```
soenkehahn, line 6: Please review this file with hidden whitespace changes!

test/spec/BuildSpec.hs
```diff
@@ -11,105 +12,166 @@ import Test.Mockery.Directory
...
+              );
+            |]
+        output <- runGarn ["build", "project.foo"]
+        exitCode output `shouldBe` ExitSuccess
+        readFile "result/build-artifact" `shouldReturn` "foo\n"
+        output <- runGarn ["build", "project.bar"]
+        exitCode output `shouldBe` ExitSuccess
+        readFile "result/build-artifact" `shouldReturn` "bar\n"
+
+      it "builds top-level packages" $ \runGarn -> do
```
soenkehahn, line 77: This wasn't required by #391, but it was easier to support this than not.

ts/internal/runner.ts
```diff
@@ -238,7 +246,7 @@ const collectProjectChecks = (
   project: Project,
 ): Record<string, Check> =>
   mapKeys(
-    (name) => `${projectName}_${name}`,
+    (name) => `${projectName}/${name}`,
```
soenkehahn, line 249: This change is unrelated but made for consistency.

ts/package.ts
```diff
@@ -27,7 +25,7 @@ export function isPackage(x: unknown): x is Package {
  */
 export function mkPackage(
   nixExpression: NixExpression,
-  description?: string,
+  description: string,
```
soenkehahn, line 28: This is a breaking change of the typescript library. I do think that it's nicer to make the description be required.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## v0.0.16
```
alexdavid, line 3:
```suggestion
## v0.0.16 (unreleased)
```

Maybe part of release process is to remove `(unreleased)` and add the next version with it to make it more obvious when it is ok to add new features to a version without checking which version is currently released?

test/spec/BuildSpec.hs
```diff
@@ -11,105 +12,166 @@ import Test.Mockery.Directory
...
+                  longer: garn.build`
+                    # this is some longer build script:
+                    bla bla bla
+                  `,
+                },
+              );
+            |]
+        output <- runGarn ["build"]
+        stderr output `shouldContain` "project.short            Builds short command"
+        stderr output `shouldContain` "project.longer           Builds # this is some longe..."
```
alexdavid, line 108: TIOLI, just felt like a change of number of spaces between the name and description shouldn't break this test
```suggestion
        stderr output `shouldMatch` "project.short \\s+ Builds short command"
        stderr output `shouldMatch` "project.longer \\s+ Builds # this is some longe\\.\\.\\."
```

(The extra spaces around `\\s+` make it a bit easier to read, and will only fail if there are fewer than 3 spaces which felt reasonable to me)

alexdavid (approved):
Looks great! I'm super excited for this!

CHANGELOG.md
soenkehahn: Yeah, that sounds right. Also, we want the changelog file on github to be explicit about what's released and what's not. For users, I mean.

PR #397, opened 2023-11-10
Add a plugin to bundle vite projects

soenkehahn:


ts/javascript/vite.ts
```diff
@@ -0,0 +1,26 @@
+import { Executable } from "../executable.ts";
+import { Package } from "../package.ts";
+import { Plugin } from "../project.ts";
+
+export const plugin: Plugin<
+  { build: Package; dev: Executable; preview: Executable },
+  { node_modules: Package }
```
jkarni, line 7: It seems unfortunate that you have to have a `node_modules` package - that doesn't seem like something end users should have to care about, and it just pollutes the API. Maybe this isn't in scope for this PR, but it feels to me like, if all we're doing in mkNpmProject is creating the node_modules dir, we *shouldn't* have it create a `Package` at all (does anyone really want to build that?). In the Environment we *do* build it, and set a var like NODE_MODULES to the location, and then below you don't need to depend on `base.node_modules`, but can instead just `export PATH=$NODE_MODULES/bin:$PATH`

ts/javascript/vite.ts
```diff
@@ -0,0 +1,26 @@
...
+  return {
+    build: base.defaultEnvironment.build`
+      export PATH=${base.node_modules}/bin:$PATH
+      vite build --outDir $out
+    `,
+    dev: base.defaultEnvironment.shell`
+      export PATH=${base.node_modules}/bin:$PATH
+      vite`,
+    preview: base.defaultEnvironment.shell`
+      export PATH=${base.node_modules}/bin:$PATH
```
jkarni, line 23: Maybe this should be set by the environment, but up to you.

jkarni (commented):
Mostly TIOLI comments

ts/javascript/vite.ts
soenkehahn: I think it should be, but not in this PR, since I think there's some other things that we want to change then. For example instead of having the initializer generate `.addExecutable("npm run foo")` we could have it put in the actual script snippet from the `package.json` file.

ts/javascript/vite.ts
```diff
@@ -0,0 +1,26 @@
+import { Executable } from "../executable.ts";
+import { Package } from "../package.ts";
+import { Plugin } from "../project.ts";
+
```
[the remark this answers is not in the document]
soenkehahn: Done in ab44005f.

ts/javascript/vite.ts
```diff
@@ -0,0 +1,26 @@
...
+export const plugin: Plugin<
+  { build: Package; dev: Executable; preview: Executable },
+  { node_modules: Package }
+> = (base) => {
+  if (base.defaultEnvironment == null) {
+    throw new Error(
+      `'vite.addBuilder' can only be added to projects with a default environment`,
```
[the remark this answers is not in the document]
soenkehahn: Done in 73321ead.

ts/javascript/vite.ts
```diff
@@ -0,0 +1,26 @@
...
+    build: base.defaultEnvironment.build`
+      export PATH=${base.node_modules}/bin:$PATH
+      vite build --outDir $out
+    `,
+    dev: base.defaultEnvironment.shell`
+      export PATH=${base.node_modules}/bin:$PATH
+      vite`,
+    preview: base.defaultEnvironment.shell`
+      export PATH=${base.node_modules}/bin:$PATH
+      vite preview`,
```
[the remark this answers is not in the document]
soenkehahn: No, we put that in automatically [here](https://github.com/garnix-io/garn/blob/73321eadcbf62437606275867426ee285966edc0/ts/executable.ts#L76). (Maybe that should change, but right now it's how it works.)

ts/project.test.ts
```diff
@@ -50,3 +56,155 @@ const _testTypeCheckingOfAddExecutableTemplate = (project: Project) => {
   // @ts-expect-error - shell should be the actual executable now, not the helper
   p.shell``;
 };
+
+describe("Project.add", () => {
+  it("allows adding fields with .add", () => {
+    const project = garn
+      .mkProject({ description: "" }, {})
+      .add((self) => ({ ...self, foo: garn.shell("echo foo") }));
```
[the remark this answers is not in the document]
soenkehahn: Done in e507566c in https://github.com/garnix-io/garn/pull/410.

ts/project.ts
```diff
@@ -15,13 +15,17 @@ import { markAsMayNotExport } from "./internal/may_not_export.ts";
...
 
-type ProjectData = {
+export type ProjectData = {
   tag: "project";
   description: string;
   defaultEnvironment?: Environment;
   defaultExecutable?: Executable;
 };
 
+export type Plugin<Additions, Dependencies = object> = (
```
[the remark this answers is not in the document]
soenkehahn: Done in 530a79ea in #410.

ts/testUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+  } catch (e) {
+    printOutput(output);
+    throw e;
+  }
+};
+
+export const runExecutable = (executable: garn.Executable): Output => {
+  const tempDir = Deno.makeTempDirSync({ prefix: "garn-test" });
+  const nixpkgsInput = nix.nixFlakeDep("nixpkgs-repo", {
+    url: "github:NixOS/nixpkgs/6fc7203e423bbf1c8f84cccf1c4818d097612566",
```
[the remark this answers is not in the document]
soenkehahn: Yes, this should've been put in `version.json` I think. I've done that in fb3d7fa0 in #410. Curious to hear what you think.

ts/javascript/vite.ts
soenkehahn: I think that's a really good idea. It would mean that we have to change how `Environment`s work. Which is something we should do, but maybe this should be deferred for now?

The details are that currently we *don't* have a way of setting an env var for `Executable`s. Just for `Package`s and `Check`s. So the solution with an env var (say `NODE_MODULES`) would only work for the `build` field, not for `dev` and `preview`. We could of course just remove those two, since this ticket is mostly about bundling. I've done that in here: dc08eca0

What do you think? Also, @alexdavid, what do you think about this? Maybe we should not merge that commit and rather re-think `Environment`s separately?

soenkehahn: @alexdavid and I brainstormed a bit about `Environment`s and how to make them better. I think we should do this after this PR.

ts/javascript/vite.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+  return {
+    build: base.defaultEnvironment.build`
+      set -eu
+
+      export PATH=node_modules/.bin:$PATH
+      ${nixRaw`pkgs.which`}/bin/which vite 2> /dev/null || \
+        (echo vite is not a dependency of the project, maybe run:
+         echo '  npm install --save-dev vite'
+          exit 1)
```
alexdavid, line 51: Maybe an if is nicer here?

```suggestion
      if ! ${nixRaw`pkgs.which`}/bin/which vite 2> /dev/null; then
        echo 'vite is not a dependency of the project, maybe run:'
        echo '  npm install --save-dev vite'
        exit 1
      fi
```

ts/javascript/vite.ts
```diff
@@ -0,0 +1,55 @@
+import { Executable } from "../executable.ts";
+import { nixRaw } from "../nix.ts";
+import { Package } from "../package.ts";
+import { Plugin } from "../project.ts";
+
+/**
+ * `Plugin` for [vite](https://vitejs.dev/) projects. This will add to your
```
alexdavid, line 7: Does the markdown link syntax work in your editor or deno doc? It doesn't have any affect in my LSP

ts/javascript/initializers.test.ts
```diff
@@ -117,9 +117,111 @@ Deno.test(
...
+      assertEquals(
+        result.makeTarget(),
+        outdent`
+          export const somepackage = garn.javascript.mkNpmProject({
+            description: "just some package",
+            src: ".",
+            nodeVersion: "18",
+          })
+            .add(garn.javascript.vite)
+            .addExecutable("dev", "npm run dev")
```
alexdavid, line 187: Did we not want to provide the `dev` script?

ts/javascript/initializers.ts
```diff
@@ -42,17 +46,28 @@ export const npmInitializer: Initializer = (dir) => {
...
-              `npm run ${name}`,
-            )})`,
-        ),
-      ].join("\n  "),
+        ...(isViteProject ? [`.add(garn.javascript.vite)`] : []),
+        ...("test" in scripts ? ['.addCheck("test", "npm run test")'] : []),
+        ...nonTestScripts
+          .filter(([name, script]) =>
+            isViteProject ? !isSuperSeededByVitePlugin(name, script) : true,
+          )
```
alexdavid, line 54: Nit, but maybe a single filter is more readable? TIOLI

```suggestion
        ...Object.entries(scripts)
          .filter(([name, script]) => {
            if (name === "test") return false;
            if (isViteProject && isSuperSeededByVitePlugin(name, script)) return false;
            return true;
          })
```

Then we can remove `nonTestScripts` above.

alexdavid (approved):

ts/javascript/vite.ts
soenkehahn: Ok, I've removed that commit.

ts/javascript/vite.ts
soenkehahn: Yeah, it does work, both in my editor and on doc.deno.land. It's a bit cumbersome to click on it in my editor, since it's a hover, but it's possible. Generally deno seems to both recommend and use markdown. (It's recommended [here](https://docs.deno.com/runtime/manual/references/contributing/style_guide#use-jsdoc-for-exported-symbols) and here's [an example](https://deno.land/std@0.206.0/toml/mod.ts?source=#L6).)

ts/javascript/initializers.test.ts
soenkehahn: No, we don't. This was back from my change discussed in https://github.com/garnix-io/garn/pull/397#discussion_r1392144122, but now it's omitted again.

PR #398, opened 2023-11-10
Fix generic helper shared namespace bugs

alexdavid:
This fixes the runtime errors when there is an executable/check/package on a project named `shell`, `check`, or `build`. We were using these helpers internally on `T` assuming that they were available, but since we share scope with user-defined executables/checks/packages we shouldn't assume this.

I think in a future PR it would be nice to also move `ProjectData` in an object nested under a `Symbol()` so that a project dep named `"description"` for example doesn't break things, but for now that seems a lot less likely to happen.

### DIFF
--- a/ts/project.test.ts
+++ b/ts/project.test.ts
@@ -2,51 +2,51 @@ import { Check } from "./check.ts";
 import { Executable } from "./executable.ts";
 import { Project } from "./project.ts";
 
-const assertIsCheck = (_c: Check) => {};
-const assertIsExecutable = (_e: Executable) => {};
+const assertTypeIsCheck = (_c: Check) => {};
+const assertTypeIsExecutable = (_e: Executable) => {};
 
-const _testAddCheck = (project: Project) => {
+const _testTypeCheckingOfAddCheck = (project: Project) => {
   const p = project
     .addExecutable("unrelated1", "true")
     .addCheck("check", "true")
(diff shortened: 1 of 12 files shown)


ts/project.test.ts
```diff
@@ -0,0 +1,52 @@
+import { Check } from "./check.ts";
+import { Executable } from "./executable.ts";
+import { Project } from "./project.ts";
+
+const assertTypeIsCheck = (_c: Check) => {};
```
soenkehahn, line 5: I wanted to make it a little bit more self-explanatory that all these definitions here are just for asserting on type checks and not being executed. I added 138ec9e9 for that. TIOLI!

ts/project.ts
```diff
@@ -137,62 +139,62 @@ export function mkProject<Deps extends Record<string, Nestable>>(
...
 const proxyEnvironmentHelpers = (): ProjectHelpers => ({
   shell(
-    this: Project,
+    this: ProjectData,
     s: TemplateStringsArray | string,
     ...args: Array<NixStrLitInterpolatable>
   ) {
-    if (this.defaultEnvironment == null) {
+    const { defaultEnvironment } = this;
+    if (defaultEnvironment == null) {
```
soenkehahn, line 147: Is this necessary?

soenkehahn (approved):

Issue #393, opened 2023-11-09
Allow to easily deploy to `gh_pages`


soenkehahn:
If I have a javascript project with a `Package` that builds the frontend bundle, it should be really easy to add an `Executable` that will deploy that built frontend bundle to the `gh-pages` branch on github.

It should *not* modify your local git checkout state in any relevant way.

Possible api:

```ts
export const website = garn.javascript
  .mkNpmProject({
    description: "The garn.io website",
    nodeVersion: "18",
    src: ".",
  })
  .addPackage("build", "npm run build ; mv build/* $out/")
  .addGhPagesDeploy("build");

  // or:
  .add(ghPagesDeploy("build", "npm run build ; mv build/* $out/"))

  // or: (@soenkehahn's personal favorite)
  .add(ghPagesDeploy("build"))
```

jkarni:
Maybe even ` .add(ghPagesDeploy())` and the name defaults to something like `deploy-gh-pages`, since most people probably don't care about customizing that, and it's not immediately obvious what the argument represents.

soenkehahn:
So the `ghPagesDeploy` function needs to somehow know, which `Package` to deploy to the `gh-pages` branch. That's what the `"build"` argument does.

jkarni:
Ah! It'd be nice if the argument were somehow of type `Package` then. Though unclear how to do that nicely (cf. #396)

alexdavid:
Yeah we spoke about something like this maybe:

```typescript
.add(ghPagesDeploy(self => self.build))
```

But ideally `ghPagesDeploy` could be made to work with all of these alternatives:
```typescript
.add(ghPagesDeploy("build"))
.add(ghPagesDeploy(self => self.build))
.add(self => ghPagesDeploy(self.build)(self))
```

soenkehahn:
Yeah, exactly. So one option that @alexdavid and I were thinking of is that
1. `Project.add` doesn't do much to make these things super convenient. Its type would just be something like `<T extends Project, U extends Project>(p: T) => U`.
2. When you write a plugin, you can tweak this as much as you want. For example `ghPagesDeploy` could be overloaded and allow parameters of type `string` (the `Package` name on the `Project`), `Package` (if you happen to have a `Package` handy), but also `(p: T) => Package` (so that you could write `.add(ghPagesDeploy(self => self.build))`). Or it could be much simpler.
[...]

PR #400, opened 2023-11-10
Add `Project.add` helper

alexdavid:


soenkehahn:
Trying this out in both #393 and #397 it seems that this solves our use-cases there. I think there's two open questions:

- Is `.add` the best name? Some alternatives are `.with`, `.modify`, `.withSelf`. Any other ideas? I kind of like both `.add` and `.with`. Although, technically, this could be used to *not* add something but somehow modify the project in another way, including *removing* things. So maybe `.add` is not the best name.
- Is there a good way to provide a type alias for these plugins? Something like `type Plugin<T extends ProjectData, U extends ProjectData> = (project: T) => U`. It'd be nice to be able to say e.g. `vite.addBuilder` has type `Plugin<{ node_modules: Package }, { viteBundle: Package }>`. (Modulo names. I don't like the name `Plugin`, any better ideas?) But toying around with that now I think that providing a type synonym that works well is not that easy. So maybe this PR should include a type synonym and have some type tests for that?

ts/project.test.ts
```diff
@@ -50,3 +56,157 @@ const _testTypeCheckingOfAddExecutableTemplate = (project: Project) => {
   // @ts-expect-error - shell should be the actual executable now, not the helper
   p.shell``;
 };
+
+describe("Project.add", () => {
+  it("allows to add fields with .add", () => {
```
jkarni, line 61:
```suggestion
  it("allows adding fields with .add", () => {
```
here and in a few places below

jkarni:
> Is .add the best name? Some alternatives are .with, .modify, .withSelf. Any other ideas? I kind of like both .add and .with. Although, technically, this could be used to not add something but somehow modify the project in another way, including removing things. So maybe .add is not the best name.

How about `apply`? It seems the most accurate.

ts/project.test.ts
soenkehahn: Done in a515a1ea.

soenkehahn:
> How about `apply`? It seems the most accurate.

We now changed the api of `.add` (and `Plugin`) so that it actually doesn't allow removing fields from a `Project`, see https://github.com/garnix-io/garn/blob/a515a1ea98f04244068b66aebda150067c40f17e/ts/project.ts#L222-L230. (Maybe it's possible by doing something weird, like setting a field to `undefined`, but it's certainly not an obvious way to use the API.) So I think the name `.add` makes a lot of sense and is consistent with the other functions that start with `add`.

soenkehahn (approved):

alexdavid (commented):
This looks good to me too. Going to merge

PR #406, opened 2023-11-14
Nix include git untracked files

jfroche:
Patch Nix so that it doesn't ignore untracked files in git anymore and
continue to ignore files if they are in a .gitignore.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -15,6 +15,22 @@
         pkgs = import "${nixpkgs}" {
           inherit system;
           config.allowUnfree = true;
+          overlays = [
+            (final: prev: {
+              nix = prev.nix.overrideAttrs
+                (finalAttrs: prevAttrs: {
+                  patches = (prevAttrs.patches or [ ]) ++ [
+                    (final.fetchpatch {
+                      name = "include-untracked-files.patch";
+                      url = "https://github.com/NixOS/nix/compare/2.17.1...jfroche:nix:feat/include-untracked-files-2.17.1.patch";
+                      excludes = [ "tests/functional/*" ];
+                      hash = "sha256-HCGYzArlHO3Q0jIfc7rVD6AfV2wBzqY7VeU5DavexRE=";
+                    })
+                  ];
+                  requiredSystemFeatures = (prevAttrs.requiredSystemFeatures or [ ]) ++ [ "big-parallel" ];
+                });
+            })
+          ];
         };
         strings = pkgs.lib.strings;
         lists = pkgs.lib.lists;
--- a/test/spec/BuildSpec.hs
+++ b/test/spec/BuildSpec.hs
@@ -4,7 +4,7 @@ module BuildSpec where
 
 import Data.String.Interpolate (i)
 import Data.String.Interpolate.Util (unindent)
-import Development.Shake (StdoutTrim (..), cmd)
+import Development.Shake (CmdOption (EchoStdout), StdoutTrim (..), cmd, cmd_)
 import System.Directory
 import System.Exit (ExitCode (..))
 import Test.Hspec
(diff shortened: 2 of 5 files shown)


src/Garn.hs
```diff
@@ -88,4 +88,4 @@ checkTarget targetConfig = case targetConfig of
   TargetConfigExecutable _ -> pure ()
 
 runNix :: (Input a, Output b) => a -> IO b
-runNix = Cradle.run "nix" nixArgs
+runNix = Cradle.run "nix" "--include-untracked-files" nixArgs
```
soenkehahn, line 91: I think this should go into `nixArgs`.

flake.nix
```diff
@@ -15,6 +15,25 @@
...
+                    (final.fetchpatch {
+                      name = "include-untracked-files.patch";
+                      url = "https://github.com/NixOS/nix/pull/9352.patch";
+                      excludes = [ "tests/functional/*" ];
+                      hash = "sha256-JPoQSXFKex6wFx2AoOuMTY0PvzdZ9R5wCqcsT+13i7k=";
+                    })
+                  ];
+                  # Set big-parallel, this allows us to build nix remotely
+                  # on machines where only big-parallel derivations are built
+                  # remotely.
```
soenkehahn, line 32: I'm tempted to delete this comment. If a reader has no idea what `big-parallel` is, this comment is not that helpful without actually googling and reading up on `big-parallel`. But for that, we don't need a comment. WDYT?

soenkehahn (commented):
I think we should have a test (maybe in `GarnSpec.hs`) that `garn` behaves this way now. Happy to pair on that.

flake.nix
```diff
@@ -15,6 +15,22 @@
...
           inherit system;
           config.allowUnfree = true;
+          overlays = [
+            (final: prev: {
+              nix = prev.nix.overrideAttrs
+                (finalAttrs: prevAttrs: {
+                  patches = (prevAttrs.patches or [ ]) ++ [
+                    (final.fetchpatch {
+                      name = "include-untracked-files.patch";
+                      url = "https://github.com/NixOS/nix/compare/2.17.1...jfroche:nix:feat/include-untracked-files-2.17.1.patch";
```
jfroche, line 25: I have rebased the pull request in the Nix repository. Due to the recent changes in the main branch, specifically the transition to the libgit2 git implementation, the PR can no longer be directly applied with our current Nix version (2.17.1).
To address this, I've initiated a branch to backport the necessary changes for version 2.17.1.

alexdavid (approved):
This looks good to me. I think we can also remove the `git add --intent-to-add flake.nix` call in `src/Garn/GarnConfig.hs:192` as well as the FAQ entry on the website at `website/src/pages/main.tsx:263-281`

jfroche:
We still get the output of the `git init` during the test, with something like this:

```
    builds packages in a git repository with untracked files [ ]Initialized empty Git repository in /tmp/nix-shell.P8XVK8/mockery-5db85674c541bbc8/.git/
    builds packages in a git repository with untracked files [✔]
 ```
 Not sure what would be the best to remove that.
 
I also see a lot of `warning: Git tree '...' is dirty` by nix during the tests, would it make sense to run nix with `--no-warn-dirty` to maybe reduce confusion for new comers ?

File: test/spec/BuildSpec.hs [hunk 4]
```diff
@@ -190,3 +190,11 @@ spec = do
           output <- runGarn ["build", "project"]
           readFile "result/build-artifact" `shouldReturn` "hello from setup\n"
           exitCode output `shouldBe` ExitSuccess
+
+      it "builds packages in a git repository with untracked files" $ \runGarn -> do
```
soenkehahn, line 194:
```suggestion
      it "includes untracked files when building packages" $ \runGarn -> do
```

soenkehahn (approved):

PR #409, opened 2023-11-14
Implement GitHub Pages plugin

alexdavid:


alexdavid:
This is ready for review. A few changes that I made that I don't feel super strongly about, but would like your thoughts on @soenkehahn:

* I moved the plugin to the root `garn` export, it felt strange that we nested it under `javascript` since you might have a static html site that you want to deploy to GitHub pages.
* I removed the `Config` type, instead opting to pass the project in directly to the first param. The thinking here is that while we may have additional options in the future, the only required thing I think we will ever need is the package, and I think this reads a bit nicer (and made the JSDoc read nicer as well since we don't have a great way of linking to the config type in deno doc):
```typescript
.add(deployGhPages(self => self.myPkg))
// and in the future:
.add(deployGhPages(self => self.myPkg, { pushTo: 'origin' }))
```
rather than
```typescript
.add(deployGhPages(self => ({ package: self.myPkg })))
// in the future:
.add(deployGhPages(self => ({ package: self.myPkg, pushTo: 'origin' })))
```
especially since I can't imagine any use case for any options needing something off of the `Project`, but let me know if you prefer the other way.

ts/deployToGhPages.ts
```diff
@@ -0,0 +1,73 @@
+import { Executable } from "./executable.ts";
+import { Package } from "./package.ts";
+import { Plugin } from "./project.ts";
+
+const ansiBold = "\\e[0;1m";
+const ansiReset = "\\e[0m";
+const ansiRedBold = "\\e[31;1m";
+
+/**
+ * A garn plugin that allows easy deployment of a package to GitHub pages.
```
soenkehahn, line 10: Maybe link to github pages documentation here?

ts/deployToGhPages.ts
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * @param pkg - The `Package` whose artifacts will be committed to the
+ * `gh-pages` branch. For convenience, this can also be a function that takes
+ * in a refrence to the project and returns the `Package`.
```
soenkehahn, line 14:
```suggestion
 * in a reference to the project and returns the `Package`.
```

soenkehahn (commented):
I tried this out on a project here: https://github.com/soenkehahn/garn-example-vite-vanilla (This is using a branch of garn with both the gh-pages plugin and the vite-builder plugin.)

When doing `garn run viteVanilla.deployToGhPages` in there, it creates a commit on `gh-pages`, but I get this output:

```
[...]
Created commit to "gh-pages" branch, but it has not been pushed yet
Run git push <remote> gh-pages:gh-pages to deploy
rm: cannot remove '/tmp/tmp.Mn58cn7bio/dst/assets/index-QFzqfbCu.js': Permission denied
rm: cannot remove '/tmp/tmp.Mn58cn7bio/dst/assets/index-s-MxoWxy.css': Permission denied
```

I'm assuming the error messages at the end are caused by copying something from the nix store to a temporary directory, which then will be read-only. And then trying to delete the temporary directory afterwards.

Also, when pushing to the gh-pages branch, github serves the files, however the app doesn't work because of mime-type issues. :( Maybe this is something that has to be fixed in the vite-builder plugin?

soenkehahn:
> Also, when pushing to the gh-pages branch, github serves the files, however the app doesn't work because of mime-type issues. :( Maybe this is something that has to be fixed in the vite-builder plugin?

Never mind. To deploy vite apps to github pages, you have to configure the `base` config variable, see https://vitejs.dev/guide/static-deploy#github-pages.

soenkehahn (approved):
Looks good, apart from my comments.

alexdavid:
> I'm assuming the error messages at the end are caused by copying something from the nix store to a temporary directory, which then will be read-only. And then trying to delete the temporary directory afterwards.

Yeah, our test package didn't output any nested directories. I updated the tests in c32d5a4 to create a nested dir to catch this in the future.

> Never mind. To deploy vite apps to github pages, you have to configure the base config variable, see https://vitejs.dev/guide/static-deploy#github-pages.

One thing to note is that this only applies to project pages, not user/organization pages.

ts/deployToGhPages.ts
alexdavid: I added a link to  https://pages.github.com/. Let me know if you think something like https://docs.github.com/en/pages/getting-started-with-github-pages/creating-a-github-pages-site would be better

PR #410, opened 2023-11-15
Improve Plugin system

soenkehahn:
This is a follow-up to #400.


ts/project.ts
```diff
@@ -22,11 +22,45 @@ export type ProjectData = {
   defaultExecutable?: Executable;
 };
 
+/**
+ * `Plugin`s automatically add one or more fields to your project. You can use
+ * existing plugins to automatically add a few useful fields without having to
```
jkarni, line 27:
```suggestion
 * existing plugins to automatically add prepackaged functionality to your `Project`.
```

(I think talk of "fields", which is quite structural, should be combined with talk of *function*.)

ts/project.ts
```diff
@@ -22,11 +22,45 @@ export type ProjectData = {
...
+ *   (
+ *     globPattern: string,
+ *   ): garn.Plugin<
+ *     { formatCheck: garn.Check; format: garn.Executable },
+ *     garn.ProjectHelpers
+ *   > =>
+ *   (p) =>
+ *     p
+ *       .withDevTools([
+ *         garn.mkPackage(garn.nix.nixRaw`pkgs.nodePackages.prettier`, "prettier"),
```
jkarni, line 45: I forget - what's the status of including `nodePackage` in nixpkgs? If it's merged, then this should be changed. If not, then reminder to change this (and the website!) when it is merged.

(Maybe cc @alexdavid)

jkarni (approved):

ts/project.ts
soenkehahn: Yup, not merged yet. Here's the PR: #382.

PR #413, opened 2023-11-16
Improve --help descriptions for vite and gh-pages

alexdavid:
The descriptions of these plugins was a bit sad, being something like `Executes "set -e ..."`. This makes the help text for projects using the Vite or GitHub pages plugins much nicer.


soenkehahn (approved):
Nice!

PR #415, opened 2023-11-17
Replace executable mechanism in mkHaskellProject

### LINKED ISSUE #395: Remove 'executable' from mkHaskellProject
Or make it optional, or a list. Haskell projects don't always have executables, and it's a confusing API.

jkarni:
Previously you *had* to define an executable, and there was only one allowed. This enables multiple ones or none, and also deduces them from the cabal file in the initializer.

You can both specify the executables with the options `executables` field (which is now also a list), or after the fact with `addCabalExecutable`. 
 

Fixes #395 in its way.

### DIFF
--- a/ts/haskell/initializers.test.ts
+++ b/ts/haskell/initializers.test.ts
@@ -55,33 +55,30 @@ Deno.test(
   },
 );
 
-Deno.test(
-  "Haskell initializer includes executables",
-  () => {
-    const tempDir = Deno.makeTempDirSync();
-    Deno.writeTextFileSync(
-      join(tempDir, "foo.cabal"),
-      `
+Deno.test("Haskell initializer includes executables", () => {
+  const tempDir = Deno.makeTempDirSync();
+  Deno.writeTextFileSync(
+    join(tempDir, "foo.cabal"),
+    `
     name: foo
     version: 0.0.1
 
     executable bar
       main-is: Main.hs
   `,
-    );
-    const result = mkHaskellProjectInitializer(tempDir);
-    assertEquals(result.tag, "ShouldRun");
-    if (result.tag === "ShouldRun") {
-      assertEquals(
-        result.makeTarget(),
-        outdent`
+  );
+  const result = mkHaskellProjectInitializer(tempDir);
+  assertEquals(result.tag, "ShouldRun");
+  if (result.tag === "ShouldRun") {
+    assertEquals(
+      result.makeTarget(),
+      outdent`
           export const foo = garn.haskell.mkHaskellProject({
             description: "",
             compiler: "ghc94",
             executables: ["bar"],
             src: "."
           })`,
-      );
-    }
+    );
   }
-)
+});
--- a/ts/haskell/initializers.test.ts
+++ b/ts/haskell/initializers.test.ts
@@ -60,12 +60,12 @@ Deno.test("Haskell initializer includes executables", () => {
   Deno.writeTextFileSync(
     join(tempDir, "foo.cabal"),
     `
-    name: foo
-    version: 0.0.1
+      name: foo
+      version: 0.0.1
 
-    executable bar
-      main-is: Main.hs
-  `,
+      executable bar
+        main-is: Main.hs
+    `,
   );
   const result = mkHaskellProjectInitializer(tempDir);
   assertEquals(result.tag, "ShouldRun");
--- a/ts/haskell/initializers.test.ts
+++ b/ts/haskell/initializers.test.ts
@@ -46,7 +46,6 @@ Deno.test(
         outdent`
           export const foo = garn.haskell.mkHaskellProject({
             description: "",
-            executable: "",
             compiler: "ghc94",
             src: "."
           })`,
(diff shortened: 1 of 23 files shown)


jkarni:
Just noticed that with the object literal syntax, the return type isn't right - it doesn't include the executables. Trying to fix that now.

ts/testUtils.ts
```diff
@@ -60,7 +60,9 @@ export const runExecutable = (
   executable: garn.Executable,
   options: { cwd?: string } = {},
 ): Output => {
-  const tempDir = Deno.makeTempDirSync({ prefix: "garn-test" });
+  const tempDir = options.cwd
+    ? options.cwd
+    : Deno.makeTempDirSync({ prefix: "garn-test" });
```
alexdavid, line 65: This is causing the failing tests since the gh-pages tests are assuming they run in an empty directory. We can make this work, but in order to do so we'll need to
* `git add flake.nix` in this function
* change expectations to include `flake.nix`, and `flake.lock` in the gh-pages test assertions
* append `-dirty` to the end of the git shas in the output.

Maybe it makes sense to have a separate option for the gh-pages tests to keep the current behavior?

ts/haskell/mod.ts
```diff
@@ -1,35 +1,72 @@
...
+ */
+export type HaskellAddenda = {
+  /**
+   * A package building the entire Haskell project
+   */
+  pkg: Package;
+  /**
+   * Make an executable from the cabal file available to garn.
+   */
+  addCabalExecutable: <T extends Project & HaskellAddenda, Name extends string>(
```
alexdavid, line 19: Maybe we want to only require `ProjectData` here in case some of the project helper methods are overridden?

```suggestion
  addCabalExecutable: <T extends ProjectData & HaskellAddenda, Name extends string>(
```

ts/haskell/initializers.ts
```diff
@@ -29,14 +29,16 @@ export const mkHaskellProjectInitializer: Initializer = (dir) => {
   }
   const parsedCabal = JSON.parse(decoder.decode(jsonParseResult.stdout));
 
+  const executables = parsedCabal.executables
+    ? `\n  executables: ${JSON.stringify(parsedCabal.executables)},`
+    : "";
```
alexdavid, line 34: In the tests you assert an empty array. Do we want to make this optional? Maybe it makes sense to always print this even if it's an empty array in the initializer?

ts/haskell/initializers.test.ts
```diff
@@ -46,11 +46,39 @@ Deno.test(
...
+  const tempDir = Deno.makeTempDirSync();
+  Deno.writeTextFileSync(
+    join(tempDir, "foo.cabal"),
+    `
+    name: foo
+    version: 0.0.1
+
+    executable bar
+      main-is: Main.hs
+  `,
```
alexdavid, line 68: Nit
```suggestion
    `
      name: foo
      version: 0.0.1

      executable bar
        main-is: Main.hs
    `,
```

ts/haskell/mod.test.ts
```diff
@@ -0,0 +1,38 @@
...
+              main-is: Main.hs
+              build-depends: base
+              default-language: Haskell2010
+          `,
+      );
+      Deno.writeTextFileSync(
+        `${path}/Main.hs`,
+        `
+            main = putStrLn "yo"
+          `,
```
alexdavid, line 24: Nit, the string above too
```suggestion
        `,
```

ts/testUtils.ts
alexdavid: Nit, if we do keep this as it is
```suggestion
  const tempDir = options.cwd ?? Deno.makeTempDirSync({ prefix: "garn-test" });
```

ts/haskell/mod.ts
```diff
@@ -1,35 +1,72 @@
...
  */
-export function mkHaskellProject(args: {
+export function mkHaskellProject<const Executables extends string[]>(args: {
   description: string;
-  executable: string;
   compiler: string;
+  executables?: Executables;
   src: string;
-}): Project & { pkg: Package } {
+}): Project & HaskellAddenda & Execs<Executables> {
```
alexdavid, line 47: Should we omit `Executables[number]` from `Project & HaskellAddenda`?

alexdavid (approved):
This looks great!

ts/testUtils.ts
jkarni: Yeah, not sure what to do. My first thought was to allow absolute paths in project sources, since then we could have the project point to wherever the source is. But I think that breaks a lot of assumptions and shouldn't be allowed in user code, at least. Ultimately I think the flake file *has* to be in the same directory as the source.

I think you're idea of a separate option makes sense.

ts/haskell/mod.ts
jkarni: Not sure I understand

ts/testUtils.ts
jkarni: Fixed. Let me know what you think.

ts/haskell/initializers.ts
alexdavid: Ah right, sorry I was confusing undefined with empty array. Does it make sense to template this out even if `parsedCabal.executables` is undefined?

ts/haskell/mod.ts
alexdavid: Yeah I didn't phrase that well. I meant something like this:

```suggestion
}): Omit<Project & HaskellAddenda, Executables[number]> & Execs<Executables> {
```

This way if someone has a Haskell executable named `check` for example, the `check` property on the `Project` is correctly typed as the Haskell executable instead of the method that we provide in `ProjectHelpers`.

ts/haskell/mod.ts
jkarni: I tried this a bit more, and the problem is, as you suggested, that then we end up with more code duplication in addExecutable, which I'd have to essentially inline.

ts/haskell/mod.ts
jkarni: Just doing that results in an error in the test:

```
error: TS2349 [ERROR]: This expression is not callable.
  Type 'Executable' has no call signatures.
        .addCabalExecutable("foo");
         ~~~~~~~~~~~~~~~~~~
    at file:///home/jkarni/garnix/garn/ts/haskell/mod.test.ts:32:10
```

By making the `executables` argument readonly, I can get it to work this way when it's explicitly provided, but not when it's omitted.

In general, I do worry this sacrifices type inference by default assuming projects are projects....

jkarni: That said, your suggestion is really nice in actually giving a type error when I add an executable named `addCabalExecutable`...

PR #425, opened 2023-11-22
Add ts based test for mkNpmProject

DavHau:
Also add getStorePath test helper to build a flake and get the store path of one of the outputs.

### DIFF
--- a/ts/javascript/mod.test.ts
+++ b/ts/javascript/mod.test.ts
@@ -1,0 +1,74 @@
+import {
+  afterEach,
+  beforeEach,
+  describe,
+  it,
+} from "https://deno.land/std@0.206.0/testing/bdd.ts";
+import * as garn from "./mod.ts";
+import { buildPackage } from "../testUtils.ts";
+import { assert } from "https://deno.land/std@0.206.0/assert/mod.ts";
+import { existsSync } from "https://deno.land/std@0.201.0/fs/exists.ts";
+
+describe("mkNpmProject", () => {
+  let tempDir: string;
+  beforeEach(() => {
+    tempDir = Deno.makeTempDirSync({ prefix: "garn-test" });
+  });
+  afterEach(() => {
+    Deno.removeSync(tempDir, { recursive: true });
+  });
+  it("produces correct store path structure", () => {
+    const packageJsonText = {
+      name: "my-npm-project",
+      version: "1.0.0",
+      description: "My npm project",
+      dependencies: {
+        typescript: "*",
+      },
+    };
+    const packageLockJsonText = {
+      name: "my-npm-project",
+      version: "1.0.0",
+      lockfileVersion: 3,
+      requires: true,
+      packages: {
+        "": {
+          name: "my-npm-project",
+          version: "1.0.0",
+          dependencies: {
+            typescript: "*",
+          },
+        },
+        "node_modules/typescript": {
+          version: "5.3.2",
+          resolved:
+            "https://registry.npmjs.org/typescript/-/typescript-5.3.2.tgz",
+          integrity:
+            "sha512-6l+RyNy7oAHDfxC4FzSJcz9vnjTKxrLpDG5M2Vu4SHRVNg6xzqZp6LYSR9zjqQTu8DU/f5xwxUdADOkbrIX2gQ==",
+          bin: {
+            tsc: "bin/tsc",
+            tsserver: "bin/tsserver",
+          },
+          engines: {
+            node: ">=14.17",
+          },
+        },
+      },
+    };
+    const packageJson = JSON.stringify(packageJsonText);
+    Deno.writeTextFileSync(`${tempDir}/package.json`, packageJson);
+    const packageLockJson = JSON.stringify(packageLockJsonText);
+    Deno.writeTextFileSync(`${tempDir}/package-lock.json`, packageLockJson);
+    const project = garn.mkNpmProject({
+      src: "./.",
+      description: "An NPM project",
+      nodeVersion: "18",
+    });
+    const output = buildPackage(project.node_modules, { dir: tempDir });
+    assert(
+      existsSync(`${output}/node_modules/typescript`),
+      `/node_modules/typescript not found in ${output}`,
+    );
+    assert(existsSync(`${output}/bin/tsc`), `/bin/tsc not found in ${output}`);
+  });
+});
--- a/ts/javascript/mod.ts
+++ b/ts/javascript/mod.ts
@@ -10,6 +10,7 @@ import {
   nixRaw,
   nixStrLit,
 } from "../nix.ts";
+import { nixpkgsInput } from "../testUtils.ts";
 
 export { plugin as vite } from "./vite.ts";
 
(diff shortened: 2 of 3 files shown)


ts/testUtils.ts
```diff
@@ -56,14 +63,37 @@ export const assertStderr = (output: Output, expected: string) => {
   }
 };
 
+export const getStorePath = (flakeDir: string, attribute: string): string => {
```
jkarni, line 66: This is a bit different than what we've been doing so far, which is that building always happens at the Haskell layer, and here we just generate statically a flake.nix file. 

If we were to continue in that direction, you can still refer to the output path, but not e.g. compare it for equality in Typescript. In particular, you could have `getStorePath : Package -> NixExpression`, for example (and similarly for `Environment` and `Check`.

Anyhow, I think this is *philosophically* a big change, so maybe it'd be helpful to know what the intended use-case is.

@soenkehahn @alexdavid are maybe more qualified to speak to all this, though.

jkarni: Ah, if this is for testing, never mind! Sorry for the noise!

ts/testUtils.ts
```diff
@@ -56,6 +63,36 @@ export const assertStderr = (output: Output, expected: string) => {
   }
 };
 
+/*
+ * Build and retrieve the store path of specific attribute of a given flake.
+ * This is useful for testing, as it allows to check the output of a specific
+ * derivation.
+ */
+export const getStorePath = (flakeDir: string, attribute: string): string => {
```
soenkehahn, line 71: I have a very similar function called `buildPackage` in #424 (in `testUtils.ts`). Overall the API of `buildPackage` is closer to how a `garn` user would interact with something of type `Package`. So `buildPackage` is what a garn user could do by running `garn build ...` in a terminal. So maybe that would make it easier to write tests that are closer to real use-cases. (Also `buildPackage` renders the flake file for you, so that's convenient.)

Do you think `buildPackage` would work for your test?

(There's some things that aren't ideal about `buildPackage`, for example I didn't dry up `nixpkgs` which you did and is a good idea. But to me it seems that `getStorePath` and `buildPackage` should be the same function.)

ts/javascript/project.test.ts
```diff
@@ -0,0 +1,31 @@
...
+import { assert } from "https://deno.land/std@0.206.0/assert/mod.ts";
+import { renderFlakeFile } from "../nix.ts";
+import { formatFlake } from "../internal/runner.ts";
+import * as path from "https://deno.land/std/fs/copy.ts";
+import { existsSync } from "https://deno.land/std@0.201.0/fs/exists.ts";
+
+describe("mkNpmProject", () => {
+  it("produces correct store path structure", () => {
+    const tempDir = Deno.makeTempDirSync({ prefix: "garn-test" });
+    path.copySync("./examples/npm-project/", `${tempDir}/npm-project`);
```
soenkehahn, line 13: I think overall it'd be nicer to not use the `examples` folder in here. Could we just write out all the needed files from the test? Maybe just a tiny example with one flat dependency?

soenkehahn (changes requested):
I think overall this is very much the right direction. Left a bunch of comments.

I do wonder how to deal with temporary directories in tests. We don't remove them (neither in the existing tests nor in your new ones). Which seems not ideal, but I'd be happy to ignore that for this PR.

ts/testUtils.ts
soenkehahn: We're merging #424 now. I hope that makes it easier to figure out whether `buildPackage` works for you.

ts/javascript/project.test.ts
```diff
@@ -0,0 +1,31 @@
...
+    Deno.chdir(`${tempDir}/npm-project`);
```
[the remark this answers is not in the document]
DavHau: I removed the `chdir`. It was unnecessary.
I also added some tear down instruction via `afterEach`. Now the temp dir will be deleted after the test.

ts/testUtils.ts
DavHau: I removed getStorePath and refactored the test to use buildPackage instead. I had to make slight modifications to some testUtils in order to get `system` into scope which is required by the expression generated by mkNpmProject.

ts/javascript/mod.ts
```diff
@@ -39,7 +39,7 @@ const fromNodeVersion = (
   const { pkg, permittedInsecurePackages } = nodeVersions[version];
   return {
     pkgs: nixRaw`
-      import "\${nixpkgs}" {
+      import "\${nixpkgs-repo}" {
```
soenkehahn, line 42: Currently we declare `nixpkgs` [here](https://github.com/garnix-io/garn/blob/966fe4114d091aff39c302d9a21f29a44d6d9303/ts/internal/runner.ts#L143) and then use that in other places (for example here). That's not ideal, since it's an implicit dependency. But why is changing it to `nixpkgs-repo` an improvement? Then it implicitly depends on someone somewhere else using `nixFlakeDep("nixpkgs-repo", ...`. And that seems even more implicit.

I think at some point we want to move to using something like `nixFlakeDep` whereever we need `nixpkgs`. But it's a little bit more involved to do that. So I wouldn't do that in this PR.

Or do you need this change for anything?

ts/javascript/mod.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    const packageJson = `
+      {
+        "name": "my-npm-project",
+        "version": "1.0.0",
+        "description": "My npm project",
+        "dependencies": {
+          "typescript": "*"
+        }
+      }
+    `;
```
soenkehahn, line 30: I think I would declare the contents of these functions as objects in typescript and then write them out with `JSON.stringify`. That way we get code formatting any syntax highlighting for the data. Also, I would inline `packageJson`. Like this:

```suggestion
    Deno.writeTextFileSync(`${tempDir}/package.json`, {
      "name": "my-npm-project",
      "version": "1.0.0",
      "description": "My npm project",
      "dependencies": {
        "typescript": "*"
      }
    });
    `;
```

(Same for `package-lock.json`.)

soenkehahn (approved):
I left some comments. If you agree with those and make the changes, feel free to merge.

ts/javascript/mod.ts
DavHau: FormatFlake introduces 'nixpkgs' via a let binding, but buildPackage doesn't. Therefore I did this change to make mkNpmProject compatible to buildPackage. As disvussed in discourse yesterday, we should probably fix this via a smarter nixFlakeDep that deduplicates. I could do this via another PR.

DavHau: I guess I can just make use of nixFlakeDep already here, then take care of deduplication later.

soenkehahn: Oh, right. I thought `buildPackage` would work without changing this, sorry.

I would maybe try to make `buildPackage` mimic what `formatFlake` does then. So that the tests are closer to what's actually going on. For example like this: f641ee33

DavHau: OK, I now made mkNpmProject use nixFlakeDep to refer to the flake input to be a bit more explicit about it.
I also added the `nixpkgs` and `system` bindings to the test helpers.

PR #429, opened 2023-11-22
initialize python integration

DavHau:
based on https://github.com/garnix-io/garn/pull/425

This initializes basic pyprojet.toml python support.

It is simplified and only supports dependencies which are in nixpkgs. It raises an error in case a requested dependency is not found in nixpkgs. 

At a later stage this should probably make use of some python2nix approach for getting the correct versions of dependencies.

### DIFF
--- a/examples/python/garn.ts
+++ b/examples/python/garn.ts
@@ -1,0 +1,7 @@
+import * as garn from "http://localhost:8777/mod.ts";
+import * as pkgs from "http://localhost:8777/nixpkgs.ts";
+
+export const someproject = garn.python.mkPythonProject({
+  src: ".",
+  pythonInterpreter: garn.python.interpreters.python310,
+});
--- a/ts/python/mod.test.ts
+++ b/ts/python/mod.test.ts
@@ -4,12 +4,21 @@ import {
   it,
 } from "https://deno.land/std@0.206.0/testing/bdd.ts";
 import * as garn from "./mod.ts";
-import { buildPackage, runCommand, runInDevShell } from "../testUtils.ts";
+import {
+  assertStderrContains,
+  buildPackage,
+  nixpkgsInput,
+  pkgs,
+  runCommand,
+  runInDevShell,
+} from "../testUtils.ts";
 import { assert } from "https://deno.land/std@0.206.0/assert/mod.ts";
 import { existsSync } from "https://deno.land/std@0.201.0/fs/exists.ts";
 import { assertStringIncludes } from "https://deno.land/std@0.206.0/assert/assert_string_includes.ts";
 import outdent from "https://deno.land/x/outdent@v0.8.0/mod.ts";
 import * as pythonInterpreters from "../internal/nixpkgs/pythonInterpreters/mod.ts";
+import * as nix from "../nix.ts";
+import { nixAttrSet } from "../nix.ts";
 
 const pyprojectToml = outdent`
   [build-system]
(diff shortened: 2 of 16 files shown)


ts/python/project.test.ts
```diff
@@ -0,0 +1,29 @@
...
+import * as garn from "./mod.ts";
+import { getStorePath, nixpkgsInput } from "../testUtils.ts";
+import { assert } from "https://deno.land/std@0.206.0/assert/mod.ts";
+import { renderFlakeFile } from "../nix.ts";
+import { formatFlake } from "../internal/runner.ts";
+import * as path from "https://deno.land/std/fs/copy.ts";
+import { existsSync } from "https://deno.land/std@0.201.0/fs/exists.ts";
+
+describe("mkPythonProject", () => {
+  it("produces correct store path structure", () => {
```
soenkehahn, line 11: I think it'd be nice to frame the tests more from the perspective of a `garn` user, e.g.:

```suggestion
  it("allows building a python project and running scripts from it", () => {
```

And then execute the script in the test and assert on its output. (`garn` users probably don't care that much about what ends up in the nix store, but they do care about being able to execute their python scripts.)

ts/python/project.test.ts
```diff
@@ -0,0 +1,29 @@
...
+    const tempDir = Deno.makeTempDirSync({ prefix: "garn-test" });
+    path.copySync("./examples/python-project/", `${tempDir}/python-project`);
```
soenkehahn, line 13: I do like that we have a python example directory, but would it be better to write out the python files here in this test-case without relying on files from `./examples`? So that the test is self-contained.

ts/python/mod.ts
```diff
@@ -0,0 +1,66 @@
...
+  const pkg: Package = mkPackage(
+    nixRaw`
+      let
+        lib = pkgs.lib;
+        python = pkgs.python3;
+        src = ${nixSource(args.src)};
+        pyproject = lib.pipe ("\${src}/pyproject.toml") [
+          builtins.readFile
+          builtins.fromTOML
+        ];
```
soenkehahn, line 35: Using `lib.pipe` for just two functions seems a bit overkill.

soenkehahn (commented):
Very nice! Left some comments.

ts/python/project.test.ts
DavHau: Removed the ./example for now and inlined the files instead. Will create the examples again once I have a test for it.

examples/python/garn.ts
```diff
@@ -0,0 +1,7 @@
+import * as garn from "http://localhost:8777/mod.ts";
+import * as pkgs from "http://localhost:8777/nixpkgs.ts";
+
+export const someproject = garn.python.mkPythonProject({
+  src: ".",
+  pythonInterpreter: garn.python.interpreters.python310,
```
jkarni, line 6: Elsewhere we have been doing only strings (e.g. "python-3.1.0", or just "3.1.0"), with the type being the union of the allowed strings. 

This makes it a bit more discoverable, at the cost of being a bit more work for us to generate those. But overall I think consistency is the main gain.

ts/python/mod.test.ts
```diff
@@ -0,0 +1,152 @@
...
+      ]
+    `;
+    Deno.writeTextFileSync(`${tempDir}/pyproject.toml`, pyprojectToml);
+    const project = garn.mkPythonProject({
+      src: "./.",
+      pythonInterpreter: pythonInterpreters.python310,
+    });
+    const pkg = project.package;
+    const flakeFile = nix.renderFlakeFile(
+      nixAttrSet({
```
jkarni, line 126: Why go through a manual `flake.nix`, and an explicit `nix build`, rather than use `garn` itself?

ts/python/mod.ts
```diff
@@ -0,0 +1,114 @@
...
+import { nixSource } from "../internal/utils.ts";
+import { nixRaw } from "../nix.ts";
+import * as pythonInterpreters from "../internal/nixpkgs/pythonInterpreters/mod.ts";
+import { parsePyprojectToml } from "./utils.ts";
+export { plugin as vite } from "../javascript/vite.ts";
+
+export const interpreters = pythonInterpreters;
+
+/**
+ * Creates a pyproject based python garn Project.
```
jkarni, line 13: Maybe a bit more documentation would be nice, especially a `@param` for pythonIntepreter. `mkHaskellProject` is a good inspiration (the other ones not so much)

ts/python/mod.ts
```diff
@@ -0,0 +1,114 @@
...
+ */
+export function mkPythonProject(args: {
+  // fields like name, version, description are read from pyproject.toml
```
jkarni, line 16: Maybe this should go in the user-facing documentation two lines above rather than here? (And the lines below I think should be removed - I think what should come from the language-specific config file and what should be specified manually is an overarching policy that we still have to settle on more deliberately?)

jkarni (commented):
Cool! Left some comments, but haven't quite finished!

examples/python/garn.ts
DavHau: I did implement it as strings first, but then thought, it would be more flexible by using an expression type instead. That way, the user could easily override the python interpreter in arbitrary ways by passing an expression.
But I understand the consistency concern. I can go for the strings for now, if that's preferred.

ts/python/mod.ts
DavHau: fixed via 56c2e3245609dc7cdd46ded683de46e33c3568fd

ts/python/mod.ts
DavHau: fixed the documentation placement via 56c2e3245609dc7cdd46ded683de46e33c3568fd

soenkehahn:
@DavHau: Feel free to merge!

Issue #427, opened 2023-11-22
Create a `callFlake` function


jkarni:
It should return a record of packages, checks, devShells, executables.

Issue #348, opened 2023-11-03
dogfooding `garn`


sol:
When I
```
git clone https://github.com/garnix-io/garn
cd garn && cabal install garn
```

and the use that version of `garn` on the tutorial then that fails because `deno` is missing:
```
$ garn enter
garn: deno: createProcess: exec: invalid argument (Bad file descriptor)
```

My naive assumption is that I should use nix while hacking on `garn` and then install it the nix-way, so that everything is wired up correctly.  Is that right?

So, I guess my next step would be to get familiar with nix Haskell development. However, even more fun would be if there were a `garn.ts` so that I could use `garn` to hack on `garn` itself (+ some rudimentary instructions on how to get started).

soenkehahn:
> My naive assumption is that I should use nix while hacking on `garn` and then install it the nix-way, so that everything is wired up correctly. Is that right?

Yes. The garn repo has a `flake.nix` file that defines a default `devShell` (you can enter it with e.g. `nix develop -Lc $SHELL`). That provides `deno` and a lot of other things. In there `cabal run garn:garn` should work. (Also, e.g. `just test` should work.)

And then you can use `nix profile install $GARN_REPO` to install a version of `garn` from disk into your nix profile (well, once we figure out #347). And that version of `garn` should find its own deno.

> So, I guess my next step would be to get familiar with nix Haskell development.

There's a few things to learn. The above `devShell` provides `cabal` (and `ghc`) and they should find all haskell dependencies.

> However, even more fun would be if there were a `garn.ts` so that I could use `garn` to hack on `garn` itself (+ some rudimentary instructions on how to get started).

Yes, that'd be great. I think there's currently a few things that we do in the `flake.nix` file that `garn` currently doesn't allow us to do easily if we switched. But overall it'd be great to "self-host" the garn project.

PR #430, opened 2023-11-22
Implement importFlake

alexdavid:
Short POC that implements #427 and is a step toward completing #348. Just threw this together to open a discussion to see if this is the direction we were all thinking about for #427, or if we wanted something different.


ts/plugins/importFlake.ts
```diff
@@ -0,0 +1,46 @@
...
+import { Plugin } from "../project.ts";
+
+export function importFlake<
+  Apps extends Record<string, string>,
+  Checks extends Record<string, string>,
+>(
+  flakeDir: string,
+  toImport: {
+    apps: Apps;
+    checks: Checks;
```
soenkehahn, line 14: Interesting. I was thinking that this would somehow create records for `Package`s, `Check`s & `Executable`s with dynamic (i.e. not statically known) keys. But this is of course kind of nice. Although, maybe sometimes you just want to import a flake and ignore type-safety? Maybe we want to allow both? So maybe this should add all the given fields to the project in a type safe way, but *also* add everything else in an unsafe way somehow?

In any case, why does `Apps` extend `Record<string, string>`? Is the key the name under which it'll be added to the project and the value is the name under which it's known in the flake file?

flake/flake.nix
```diff
@@ -0,0 +1,224 @@
+{
```
soenkehahn, line 1: Maybe put this files into `.garn/legacy-flake.nix` if possible, otherwise `.garn/legacy/flake.nix`?

jkarni: Maybe we should have two dirs, data and gen, to distinguish things we use from things we generate

jkarni: Maybe we should have two dirs, data and gen, to distinguish things we use from things we generate

ts/plugins/importFlake.ts
jkarni: I also hadn't thought of it, but I like the explicit and typed solution.

Not sure I like it being a `Plugin`, though. The fact that the `project` argument is ignored is, in my mind, evidence that it shouldn't be a plugin. "Lifting" it to a `Plugin` is always trivial. And there are other patterns of usage (such as just picking a few entries to add to one project, and a few to another) that also make sense.

If it's *not* a plugin, I think it also doesn't make sense to do name remapping in the import itself. I'm imagining usage being something like:

```typescript
const flake = importFlake({
  apps: ["app1", "app2"],
  checks: ["formatting", "tests"],
  packages: ["backend"] 
})

export const foo = someproj.add((_) => { 
  format: flake.checks.formatting, 
  server: flakes.packages.backend
 })

export const foo = someproj.add((_) => { 
  ci: flake.checks.tests
  devServer: flake.checks.app1
 })
```


But the more important message: this is overall WAY COOL!

soenkehahn: > But the more important message: this is overall WAY COOL!

Agreed!

I think one use-case that we should definitely support is to get *all* `Check`s, `Package`s, `Executable`s & `Environment`s from the imported flake file and allow to *somehow* expose them to `garn`. If someone ends up adding e.g. a `Check` to an imported flake file it'd be really sad if that `Check` got silently ignored on e.g. CI.

jkarni: Time for a `noUnusedFlakeExports` check/plugin :)

soenkehahn: Here's an idea for a possible api:

```ts
// allChecks is a check that passes if all checks from the subflake pass.
// allPackages is a derivation where all packages from the subflake are in subdirectories.
export const subFlake: { allChecks: garn.Check, allPackages: garn.Package} = importFromFlake(".garn/flake.nix");

// Here we promise that `whatever` exists.
// This will cause a runtime error not when executing the garn file, but only when trying to run the `Executable`.
const whatever: garn.Executable = subFlake.getExecutable("whatever");

const someCheck: garn.Check = subFlake.getCheck("someCheck");
```

This API would work without calling e.g. `nix flake show --json` when executing the garn file.

soenkehahn: Although calling `nix flake show --json` may be the right call here. It'll give much better help messages. And tab-completion may get slower, but also would be more accurate.

alexdavid:
Updated per discussion above. Let me know what you think! This is still in draft since I haven't pulled anything except the fileserver and checks.

flake/flake.nix
alexdavid: The dogfooding has been moved to #469

ts/importFlake.test.ts
```diff
@@ -0,0 +1,226 @@
...
+          main_pkg = pkgs.runCommand "create-some-files" {} ''
+            mkdir $out
+            touch $out/foo $out/bar
+          '';
+        };
+      }`);
+      const flake = importFlake("./to-import");
+      const exe = garn.shell`ls ${flake.getPackage("main_pkg")}`;
+      const output = assertSuccess(runExecutable(exe, { cwd: tempDir }));
+      assertStdout(output, "bar\nfoo\n");
```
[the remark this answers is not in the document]
soenkehahn: Some for the other tests that use `getPackage`.

ts/importFlake.ts
```diff
@@ -0,0 +1,63 @@
+import { Check } from "./check.ts";
+import { Environment, mkEnvironment } from "./environment.ts";
+import { Executable, mkExecutable } from "./executable.ts";
+import { NixExpression, nixFlakeDep, nixRaw, nixStrLit } from "./nix.ts";
+import { mkPackage, Package } from "./package.ts";
+
+function getFlake(flakeDir: string): NixExpression {
+  if (flakeDir.match(/^[.][.]?[/]/)) {
```
soenkehahn, line 8: This matches relative and absolute paths?

Btw, there's also lots of other flake url syntax options, see [here](https://nixos.org/manual/nix/unstable/command-ref/new-cli/nix3-flake#url-like-syntax) and [here](https://nixos.org/manual/nix/unstable/command-ref/new-cli/nix3-flake#types) lists all possible types. So e.g. `file:...` and `path:...` are valid flake urls. I don't think we have to support any of this besides what's currently implemented. But 1. I wanted you to know those exist and 2. maybe we should think about what happens with those right now.

ts/importFlake.ts
```diff
@@ -0,0 +1,63 @@
...
+      url: "github:divnix/call-flake",
+    });
+    return nixRaw`(${callFlake} ${nixRaw(flakeDir)})`;
+  }
+  return nixFlakeDep(flakeDir.replaceAll(/[^a-zA-Z0-9]/g, "-"), {
+    url: flakeDir,
+  });
+}
+
+export function importFlake(flakeDir: string): {
```
soenkehahn, line 19:
```suggestion
export function importFlake(flakeUrl: string): {
```
Since this may be a github flake url.

ts/importFlake.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    allPackages: mkPackage(
+      nixRaw`pkgs.linkFarm "all-packages" ${flake}.packages.\${system}`,
+      `All packages from flake.nix in ${flakeDir}`,
+    ),
+
+    getApp: (appName) =>
+      mkExecutable(
+        nixRaw`${flake}.apps.\${system}.${nixStrLit(appName)}.program`,
+        `Execute ${appName} from flake.nix in ${flakeDir}`,
```
soenkehahn, line 42: Should we make sure that -- in case an app with name `appName` doesn't exist in the given flake -- we get a nice error message? One that mentions `getApp` by name? (Same for the other potentially failing methods.)

soenkehahn:
Forgot to mention that I'm very excited about this PR! :)

ts/importFlake.ts
alexdavid: This matches relative paths only at the moment. Do you think we should support absolute paths? One thing I realized in testing is that the only thing that would break when using a flake input vs callFlake is parent path imports. Maybe we can experiment with `file:...` and `path:...` tomorrow and see how they interact with both flake inputs and callFlake

soenkehahn (approved):
LGTM.

PR #436, opened 2023-11-24
Refactor Environments

soenkehahn:
This PR refactors how `Environment`s set up their build sandboxes to make
source files available.

### DIFF
--- a/ts/environment.ts
+++ b/ts/environment.ts
@@ -7,6 +7,7 @@ import {
   nixList,
   nixRaw,
   nixStrLit,
+  unlinesNixStrings,
 } from "./nix.ts";
 import { Package, mkShellPackage } from "./package.ts";
 
@@ -23,7 +24,7 @@ import { Package, mkShellPackage } from "./package.ts";
 export type Environment = {
   tag: "environment";
   nixExpression: NixExpression;
-  setup: NixExpression;
+  buildSandboxSetup: Array<NixExpression>;
 
   /**
    * Creates a new environment based on this one that includes the specified nix packages.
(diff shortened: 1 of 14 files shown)


ts/environment.ts
```diff
@@ -152,17 +163,18 @@ export function mkEnvironment(
...
+    chmod -R u+rwX .
   `;
   return {
     tag: "environment",
     nixExpression: args.nixExpression ?? nixRaw`pkgs.mkShell {}`,
-    setup: nixStrLit`
-      ${args.src != null ? copySource(args.src) : ""}
-      ${args.setup || nixStrLit``}
-    `,
+    buildSandboxSetup: [
```
jkarni, line 172: Maybe change back to just string? TIOLI

ts/environment.ts
```diff
@@ -152,17 +163,18 @@ export function mkEnvironment(
 ): Environment {
   const copySource = (src: string) => nixStrLit`
     echo copying source
-    cp -r ${nixSource(src)} src
-    chmod -R u+rwX src
-    cd src
+    cp -r ${nixSource(src)}/. .
```
jkarni, line 166: The docstring above is outdated (missing `src` param, and has `mkEnvironment` not take a record)

jkarni, line 166: Maybe rename argument above to something like `sandboxSetup`?

ts/environment.ts
jkarni, line 172: Could also be renamed to `sandboxSetup`, but TIOLI

ts/environment.ts
```diff
@@ -55,6 +56,16 @@ export type Environment = {
   ): Package;
 };
 
+export const buildSandboxScript = (
```
jkarni, line 59:
```suggestion
export const sandboxScript = (
```

jkarni (approved):
Some TIOLIs, but approved (modulo the docstring actually needing to change)

PR #437, opened 2023-11-24
Fix double-dash subcommand parsing

jkarni:
Previously you had to pass `--` *twice* in order to get it to an underlying executable. This PR fixes that by inlining subparsers.

A further advantage of this change is that you can now mix global and local (subcommand-specific) options. It seems like the old behavior was only there for backwards-compatibility reasons (see https://github.com/pcapriotti/optparse-applicative/issues/433#issuecomment-892526113).


soenkehahn (approved):
Nice!

PR #438, opened 2023-11-24
Run args

jkarni:
Explicit shell/run arguments
    
 Previously we just tacked on $@ to the end of the shell script. But often this is completely the wrong place for it, and for basically  anything that is a proper shell script (e.g. editGarnConfig) rather than single command, makes it impossible to correctly pass arguments down.
    
 This PR changes behavior so that you have to explicitly use $@/$*. This is exactly like any bash script, so I don't think it should be too confusing.

Stacked on #437

### DIFF
--- a/website/src/docs/concepts.mdx
+++ b/website/src/docs/concepts.mdx
@@ -89,6 +89,19 @@ export const hello =
 
 You can run `Executable`s with `garn run`.
 
+If you would like `garn run` to pass arguments to your executable, use `"$@"`
+where you'd like to receive them. For examples
+
+```typescript
+import * as garn from "https://garn.io/ts/v0.0.16/mod.ts";
+
+export const hello =
+  garn.emptyEnvironment.shell('printf "My first two arguments were %s and %s" "$@"');
+```
+
+This works exactly like bash-script arguments, so it helps to know the
+differences between `"$@"`, `$@`, and `$*`.
+
 ## Packages
 
 (See also the [API docs](https://doc.deno.land/https://garn.io/ts/v0.0.16/mod.ts/~/Package).)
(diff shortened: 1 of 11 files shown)


website/src/docs/concepts.mdx
```diff
@@ -89,6 +89,19 @@ export const hello =
 
 You can run `Executable`s with `garn run`.
 
+If you would like `garn run` to pass arguments to your executable, use `"$@"`
+where you'd like to receive them. For examples
```
soenkehahn, line 93:
```suggestion
where you'd like to receive them. For example:
```

website/src/docs/concepts.mdx
```diff
@@ -89,6 +89,19 @@ export const hello =
...
+
+```typescript
+import * as garn from "https://garn.io/ts/v0.0.16/mod.ts";
+
+export const hello =
+  garn.emptyEnvironment.shell('printf "My first two arguments were %s and %s" "$@"');
```
soenkehahn, line 99:
```suggestion
  garn.shell('printf "My first two arguments were %s and %s" "$@"');
```

That should work, right?

website/src/docs/concepts.mdx
```diff
@@ -89,6 +89,19 @@ export const hello =
...
+```
+
+This works exactly like bash-script arguments, so it helps to know the
+differences between `"$@"`, `$@`, and `$*`.
```
soenkehahn, line 103: This feels a bit ... condescending? So maybe
```suggestion
This works exactly like bash-script arguments. For the exact behavior, see `man bash`, section `PARAMETERS`.
```

soenkehahn (approved):
Yeah, I think this is probably better than the current behavior. Although it's a breaking change and has some downsides, e.g. something like `garn.shell("eslint")` doesn't allow passing in arguments anymore. Still, I think an improvement.

cc: @alexdavid

PR #441, opened 2023-11-24
Make `edit` a top-level command

jkarni:
It was hard to discover before, involved more work to setup, and could break if you incorrectly edited the file.


nix/editor.nix
```diff
@@ -0,0 +1,77 @@
+let
+  pkgs = import
+    (builtins.fetchTarball {
+      url = "https://github.com/NixOS/nixpkgs/archive/a78de4f116747585ebdb6f3d725268a7e068c554.tar.gz";
```
jkarni, line 4: The reason not to use the top-level nixpkgs is that every time we update that (we should hopefully be ~ once a week?) this might need to be rebuild for all users, and we probably want a slower cadence then that.

src/Garn/Edit.hs
```diff
@@ -0,0 +1,11 @@
...
+
+import Cradle (StdoutTrimmed (..), run, run_)
+import Data.String.Conversions (cs)
+import Paths_garn (getDataFileName)
+
+editGarnTs :: [String] -> IO ()
+editGarnTs args = do
+  editorNix <- getDataFileName "nix/editor.nix"
+  StdoutTrimmed path <- run "nix-build" [cs editorNix, "--no-out-link"]
+  run_ (cs path <> "/bin/vscodium") ("garn.ts" : args)
```
soenkehahn, line 11: I think in the future we probably want to implement something like `garn edit --vim-style-vscodium` or `garn edit --vim` or `garn edit --emacs`. Is that just going to be a breaking change then? What's the use-case for allowing to pass arguments to `vscodium`?

src/Garn/Optparse.hs
```diff
@@ -139,6 +149,15 @@ withoutGarnTsParser =
...
       ]
 
+alwaysParser :: Parser AlwaysCommand
+alwaysParser =
+  subparser $ OA.command "edit" (info (Edit <$> argvParser) (progDesc desc))
+  where
+    argvParser :: Parser [String]
+    argvParser = many $ strArgument $ metavar "...args"
+
+    desc = "Edit garn.ts in VSCodium with extensions installed"
```
soenkehahn, line 159: This is a bit unspecific about what extensions are installed. Maybe:

```suggestion
    desc = "Edit garn.ts in VSCodium with Deno integration set up"
```

test/spec/EditSpec.hs
```diff
@@ -0,0 +1,23 @@
...
+spec :: Spec
+spec =
+  describe "edit" $ around onTestFailureLogger $ do
+    repoDir <- runIO getCurrentDirectory
+    around_
+      ( withModifiedEnvironment [("NIX_CONFIG", "experimental-features =")]
+          . inTempDirectory
+      )
+      $ do
+        it "starts VSCode with the deno extension" $ \onTestFailureLog -> do
```
soenkehahn, line 20:
```suggestion
        it "starts VSCodium with the deno extension" $ \onTestFailureLog -> do
```

test/spec/EditSpec.hs
```diff
@@ -0,0 +1,23 @@
...
+          output <- runGarn ["edit", "--", "--list-extensions"] "" repoDir Nothing
+          onTestFailureLog output
+          stdout output `shouldBe` "denoland.vscode-deno\n"
```
soenkehahn, line 23: Maybe add a test that `garn edit` works even when the `garn.ts` file has type errors? (Spoiler, it doesn't.)

jkarni: Very nice catch!

src/Garn/Edit.hs
jkarni: The main use-case is writing tests! But currently you still need to pass `--`, so it would still be compatible with `--vim` or whatever. (I have a hard time knowing what 'breaking change' in this context means though, and whether it matters.)

We could have e.g. an environment variable instead, that corresponds to the arguments passed. But that feels quite hacky too (and also `cradle` doesn't support env variables yet.) Open to other suggestions though!

nix/editor.nix
```diff
@@ -0,0 +1,77 @@
...
+    export HOME=$TEMP_DIR
+    export XDG_CONFIG_HOME=$TEMP_DIR/.config
+    export XDG_CACHE_HOME=$TEMP_DIR/.cache
+
+    ${vscodium}/bin/codium --new-window --disable-workspace-trust ./garn.ts --wait "$@"
+
+    rm -rf $TEMP_DIR
+  '';
+in
+editor
```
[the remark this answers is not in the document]
jkarni: Nothing in the logic is *supposed* to have changed. That said, I did inline one or two things since it seemed sensible to.

src/Garn/Edit.hs
soenkehahn: Oh, right, you have to use `--`. That's a bit surprising, but I think fine.

nix/editor.nix
jkarni: Good point.

It seems like VSCodium uses it's own deno version. So I guess it's between using the nixpkgs from flake.nix and getting matching denos, or using a different one and getting less recompilation. (I don't feel strongly one way or the other.)

src/Garn/Optparse.hs
```diff
@@ -137,16 +149,16 @@ withGarnTsParser targets =
         | (cmd, desc, runner) <- withGarnTsCommandInfo
       ]
 
-withouGarnTsCommandInfo :: [(String, String, Parser WithoutGarnTsCommand)]
-withouGarnTsCommandInfo =
+withoutGarnTsCommandInfo :: [(String, String, Parser WithoutGarnTsCommand)]
+withoutGarnTsCommandInfo =
```
soenkehahn, line 153: :birthday:

src/Garn/Optparse.hs
```diff
@@ -165,3 +198,6 @@ data CheckCommandOptions
   = Qualified CommandOptions
   | Unqualified
   deriving stock (Eq, Show)
+
+data HiddenParser = Hidden | NotHidden
```
soenkehahn, line 202: `NotHidden` is not used. I think I don't understand why `edit` is showing up as a subcommand. Maybe we should have tests for the behavior we want?

src/Garn/Optparse.hs
```diff
@@ -25,19 +28,32 @@ getOpts oType =
...
-      WithoutGarnTs -> WithoutGarnTsOpts <$> withoutGarnTsParser
+      WithGarnTs garnConfig ->
+        (WithGarnTsOpts garnConfig <$> withGarnTsParser (targets garnConfig))
+          <|> (AlwaysAvailableOpts <$> alwaysParser Hidden)
+      WithoutGarnTs ->
+        (WithoutGarnTsOpts <$> withoutGarnTsParser)
+          <|> (AlwaysAvailableOpts <$> alwaysParser Hidden)
+      BrokenGarnTs err ->
+        (AlwaysAvailableOpts <$> alwaysParser Hidden)
+          <|> throw (Garn.Errors.UserError $ errorMessage err)
```
soenkehahn, line 56: Regarding throwing a pure exception here and tab-completion: Currently syntax errors in the garn file allow to run `garn edit`, but tab-completion doesn't work nicely, it just outputs the garn deno error. Here's a test that fails and proves that:

```haskell
        it "tab-completes edit when garn.ts has syntax error" $ \onTestFailureLog -> do
          writeFile "garn.ts" "{ syntax err"
          output <- runGarn ["--bash-completion-index", "0"] "" repoDir Nothing
          onTestFailureLog output
          stdout output
            `shouldBe` unindent
              [i|
                edit
                -h
                --help
                --version
              |]
```

I didn't find a very easy way to avoid that. Turning the main parser into a `Parser (Either UserError Options)` fixes the issues with tab-completion, but means that when running other commands (e.g. `garn run`), it'll *not* output the deno error, but instead a cli help message saying that `run` is unavailable, but `edit` is allowed.

So -- unless we find a solution quickly -- I'm inclined to merge this as is, since this PR doesn't make things worse.

test/spec/RunSpec.hs
```diff
@@ -55,26 +55,36 @@ spec = do
...
       output <- runGarn ["run", "main"]
       stdout output `shouldBe` "Hello, world!\n"
       exitCode output `shouldBe` ExitSuccess
 
+    it "fails informatively on syntax errors" $ \runGarn -> do
+      writeFile
+        "garn.ts"
+        [i|
+          {askd "shyntax err"
+        |]
```
soenkehahn, line 73:
```suggestion
      writeFile "garn.ts" "{syntax err"
```

website/src/docs/getting_started.mdx
```diff
@@ -97,25 +97,11 @@ nix --extra-experimental-features 'nix-command flakes' profile install -L github
...
-```typescript
-export const edit = garn.editGarnConfig;
-```
-
-And then run it with `garn run edit`.
-
-This is a very easy way to get up and running with editing `garn.ts` files. The
-disadvantage is that the editor isn't your normal configured editor. So you
-might consider installing the Deno LSP and configuring your editor:
+<Garn /> provides an `edit` command that downloads and starts a VSCodium configured with Deno LSP. It is started with it's own configuration and data directories, so as to not clobber any configuration you may already have.
```
soenkehahn, line 104:
```suggestion
<Garn /> provides an `edit` command that downloads and starts a VSCodium configured with Deno LSP. It is started with it's own configuration and data directories, so as to not clobber any VSCodium or VSCode configuration you may already have.
```

website/src/pages/main.tsx
```diff
@@ -200,38 +200,22 @@ $ garn run backend.codegen `}
...
           <h3>
-            <code>editGarnConfig</code>
+            <code>garn edit</code>
           </h3>
           <p>
-            <Garn /> offers an <code>Executable</code> called{" "}
-            <code>editGarnConfig</code>. It will spin up a{" "}
+            <Garn /> can download and configure{" "}
+            <a href="https://vscodium.com/">vscodium</a> for you &mdash; just
+            type <code>garn edit</code> It will spin up a{" "}
```
soenkehahn, line 213:
```suggestion
            type <code>garn edit</code>. It will spin up a{" "}
```

README.md
```diff
@@ -88,6 +88,10 @@ checks run in a sandbox. The downside of sandboxing is that these checks won't
 have access to the internet. The upside is that they'll be (almost) perfectly
 reproducible.
 
+### `garn edit`
+
+`garn edit` will start (and, if necessary, download) VSCodium, with Deno LSP configured for you. It will open `garn.ts` in the current directory. This won't clobber any of your existing VSCode/VSCodium configuration and data.
```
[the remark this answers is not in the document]
jkarni: I feel like `and` reads more naturally here...

nix/editor.nix
jkarni: There are some complications with that. If it's in the `flake` file, then we either make it a dependency of `garn` (downside: everyone has to download it when installing garn, even if they never use it), or we have to reference the entire directory from `garn` (basically, include the entire repo in `data-files`) so that the references of `flake.nix` work.

If we make it a separate file, then we have to use `nixpkgsInput` and hard code the sha. This is probably the best option, but if it's okay I'm going to leave it for later so I can get the release out.

src/Garn/Optparse.hs
jkarni: The original purpose of this was because: if `edit` is not hidden, instead of `garn COMMAND` in the help text, we get `garn (COMMAND | COMMAND)`. My thought was that if it was not hidden, then *when garn.ts is invalid*, we would get just `garn`, not `garn COMMAND`. But currently the help text doesn't even show up there because the `garn.ts` error takes precedence, so it doesn't really matter.

PR #442, opened 2023-11-24
Add TestUtils.withRunGarn and use it in RunSpec

soenkehahn:


test/spec/RunSpec.hs
```diff
@@ -9,265 +9,243 @@ import Data.String.Interpolate.Util (unindent)
...
+        [i|
+          import * as garn from "#{repoDir}/ts/mod.ts"
+
+          export const main = garn.shell("echo foobarbaz");
+        |]
+      output <- runGarn ["run", "main"]
+      stdout output `shouldBe` "foobarbaz\n"
+      exitCode output `shouldBe` ExitSuccess
+
+    it "propagates the exit status of the child process" $ \runGarn -> do
```
jkarni, line 42: I'm assuming the only change here is indentation?

test/spec/BuildSpec.hs
```diff
@@ -8,32 +8,13 @@ import Development.Shake (CmdOption (EchoStdout), StdoutTrim (..), cmd, cmd_)
...
-          test runGarn'
-      )
-    $ spec
-
 spec :: Spec
 spec = do
   describe "build" $ do
     repoDir <- runIO getCurrentDirectory
-    wrap $ do
+    withRunGarn $ do
```
jkarni, line 17: Maybe move this to immediately after `describe "build"` for consistency with the file below, or vice-versa? Definitely TIOLI, though

jkarni (approved):
The diff is a bit ugly, but I'm assuming the only thing changed is the obvious refactoring of `withRunGarn`, and using it? If so, LGTM

test/spec/BuildSpec.hs
soenkehahn: Yup, done.

test/spec/RunSpec.hs
soenkehahn: Well, and removing the call to `onTestFailureLog` and calling the passed in `runGarn`. But yeah, it's a very mechanical refactoring.

soenkehahn:
> The diff is a bit ugly, but I'm assuming the only thing changed is the obvious refactoring of `withRunGarn`, and using it? If so, LGTM

Yeah, exactly.

PR #443, opened 2023-11-25
Allow entering top-level environments with `garn enter`

soenkehahn:
Maybe easier to review commit by commit.


CHANGELOG.md
```diff
@@ -3,8 +3,11 @@
 ## v0.0.17 (unreleased)
 
 - Fix locale warnings on MacOS
+
 - Make `executables` a list, and optional, in `mkHaskellPackage`
+
```
jkarni, line 8: I've been meaning to ask - what's the deal with spaces in the changelog? In the unreleased version we have it, but not in the others? Why not be consistent throughout?

soenkehahn: I'm not sure. This was my attempt at just making in consistent within this section, without having to go back to older sections and change all of them. But yeah, I think I'll just remove all newlines like this.

soenkehahn: Done in 8ab84b72.

test/spec/EnterSpec.hs
```diff
@@ -215,3 +215,81 @@ spec = do
...
+                  chmod +x $out/bin/test-tool
+                `);
+
+                export const topLevelEnv = garn.emptyEnvironment.withDevTools([testTool]);
+              |]
+            output <- runGarn ["enter", "topLevelEnv"] "test-tool\nexit\n" repoDir Nothing
+            onTestFailureLog output
+            stdout output `shouldBe` "test tool output\n"
+
+          it "does not show a description for top-level environments" $ \onTestFailureLog -> do
```
jkarni, line 238: Maybe
```suggestion
          it "shows no description for environments that have no description" $ \onTestFailureLog -> do
```

test/spec/EnterSpec.hs
```diff
@@ -215,3 +215,81 @@ spec = do
...
+
+                export const topLevelEnv = garn.emptyEnvironment;
+              |]
+            output <- runGarn ["enter", "--help"] "" repoDir Nothing
+            onTestFailureLog output
+            stdout output
+              `shouldMatch` unindent
+                [i|
+                  Available commands:
+                    topLevelEnv[ ]*
```
jkarni, line 252: Wouldn't this match even things that have description? Like maybe a `$` is needed at the end?

ts/javascript/mod.ts
```diff
@@ -94,7 +93,6 @@ export function mkNpmProject(args: {
       defaultEnvironment: devShell,
     },
     {
-      devShell,
```
jkarni, line 97: Maybe worth mentioning this change in the changelog?

test/spec/EnterSpec.hs
```diff
@@ -215,3 +215,81 @@ spec = do
...
+            output <- runGarn ["enter", "--help"] "" repoDir Nothing
+            onTestFailureLog output
+            stdout output
+              `shouldMatch` unindent
+                [i|
+                  Available commands:
+                    topLevelEnv[ ]*my test environment
+                |]
+
+          it "does not show (or allow entering) the default environment of projects" $ \onTestFailureLog -> do
```
jkarni, line 275: I imagine the point is that you can just enter `p` directly? If so, shouldn't there be a test for that? If not, then why do we want this?

test/spec/EnterSpec.hs
soenkehahn: Turns out both works as intended. But with the `$` it's more obvious, so I'll change it.

test/spec/EnterSpec.hs
soenkehahn: The changes in this PR meant that the `defaultEnvironment` of projects would 1. show up in the help as e.g. `p.defaultEnvironment` and 2. you could enter it fully qualified, e.g. with `garn enter p.defaultEnvironment`. I don't think that's useful, since you can just enter the project instead (with `garn enter p`). And we already have tests for that. (For example [here](https://github.com/garnix-io/garn/blob/0e61454d1506bb95a4d3a61eb4791cf82b24bfdc/test/spec/EnterSpec.hs#L29-L48).)

So this test here is only about making sure that 1. `p.defaultEnvironment` doesn't show up in the help output and 2. that `garn enter p.defaultEnvironment` doesn't work.

Maybe the test name should be better to clarify? Or do you still think we should have yet another test for entering the default environment of projects?

ts/javascript/mod.ts
soenkehahn: Done in 4a6d1470.

test/spec/EnterSpec.hs
jkarni: No, you're right. I miss the existing test!

jkarni (approved):

PR #449, opened 2023-11-25
Add `addToSetup` and helpers to `Environment`

soenkehahn:
This PR changes how `Environment`s set up scripts, both in the build sandboxes and outside of them. Most importantly it allows to *have* a setup for `Executable`s and `garn enter`. This would allow us to experiment with e.g. modifying the `$PATH` to add `node_modules/.bin` and similar things.

I'm not sure this is the best API, so ideas for improvements are very welcome. Also, I think the naming of everything could be better and more consistent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,8 @@
 - Add `garn.callFlake` helper to allow importing flake files by local path.
 - Add `garn.importFlake` helper to allow importing flake files by url, e.g. from GitHub.
 
-- Add `Environment.addToSetup` and `Environment.addToSandboxSetup`, making setups for `Environment`s more flexible.
-- Remove `sandboxSetup` parameter from `mkEnvironment`, use `addtoSandboxSetup` instead.
+- Remove `sandboxSetup` parameter from `mkEnvironment`, add lower-level
+  `addToSetup` helper function to use instead.
 
 ## v0.0.18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
 - Add `garn.callFlake` helper to allow importing flake files by local path.
 - Add `garn.importFlake` helper to allow importing flake files by url, e.g. from GitHub.
 
+- Add `Environment.addToSetup` and `Environment.addToSandboxSetup`, making setups for `Environment`s more flexible.
+- Remove `sandboxSetup` parameter from `mkEnvironment`, use `addtoSandboxSetup` instead.
+
 ## v0.0.18
 
 - Allow entering top-level `Environment`s with `garn enter`.
--- a/test/spec/BuildSpec.hs
+++ b/test/spec/BuildSpec.hs
@@ -96,7 +96,7 @@ spec = do
             import { addToSetup } from "#{repoDir}/ts/environment.ts"
 
             const env = addToSetup(
-              "sandbox",
+              "sandboxed",
               garn.mkEnvironment(),
               garn.nix.nixStrLit`
                 mkdir dist
@@ -165,7 +165,7 @@ spec = do
               { description: "" },
               {
                 package: addToSetup(
-                    "sandbox",
+                    "sandboxed",
                     garn.mkEnvironment(),
                     nix.nixStrLit`SETUP_VAR="hello from setup"`,
                   ).build("echo $SETUP_VAR > $out/build-artifact"),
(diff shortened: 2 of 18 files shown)


ts/environment.ts
```diff
@@ -24,16 +24,42 @@ import { Package, mkShellPackage } from "./package.ts";
...
+   * most common example for how the `Environment` for these sandboxes works
+   * differently is, that we have to *copy* the source files into the sandbox,
+   * to allow `Check`s and `Package`s to access them. Whereas with `Executable`s
+   * and `garn enter`, your source files are already available normally through
+   * the file system.)
+   */
+  addToSandboxSetup(snippet: string | NixExpression): Environment;
+  /**
+   * Creates a new environment based on this one that includes the specified nix
+   * packages.
```
soenkehahn, line 62: Here we add `addToSetup` and `addToSandboxSetup` to the `Environment` type. I think the names of those could be better, but I hope the jsdocs explain well, what they do. Any better ideas for naming these?

ts/environment.ts
```diff
@@ -62,11 +88,24 @@ export type Environment = {
...
 export const sandboxScript = (
   env: Environment,
   script: NixExpression,
 ): NixExpression =>
-  unlinesNixStrings([nixStrLit`mkdir -p $out`, env.sandboxSetup, script]);
+  joinNixStrings("\n", [
+    nixStrLit`mkdir -p $out`,
+    ...env.setup.map((x) => x.snippet),
+    script,
+  ]);
```
soenkehahn, line 108: Here are two helpers, `commonScript` and `sandboxScript` that are used to create scripts that then run in either a build sandbox or outside of a build sandbox. Again, are there better ideas for names?

ts/nix.ts
```diff
@@ -195,13 +195,16 @@ export function nixStrLit(
...
+  separator: string,
+  list: Array<NixExpression>,
+): NixExpression {
   return {
     [__nixExpressionTag]: null,
     type: "strLit",
     str: {
       initial: "",
-      rest: list.map((expr) => [expr, "\n"]),
+      rest: list.map((expr, i) => [expr, i < list.length - 1 ? separator : ""]),
```
alexdavid, line 207: I ended up doing something very similar, except for `NixExpression`s vs `NixStrLit`s here: https://github.com/garnix-io/garn/pull/430/files#diff-fb4f9f2ab46707d603912af39f9d791c92eb585508b43f11b86e588c7c88513eR81-R86

I wonder if it makes sense to attempt to consolidate these somehow (when one of our PRs makes it in) or if the fact one operates on raw nix expressions and the other operates on string literals means they stay separate.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## v0.0.19 (unreleased)
 
+- Add `Environment.addToSetup` and `Environment.addToSandboxSetup`, making setups for `Environment`s more flexible.
```
alexdavid, line 5: Seeing as we discussed making these intended for internal use should we omit this changelog entry?

test/spec/BuildSpec.hs
```diff
@@ -94,17 +94,16 @@ spec = do
...
-                echo build-content > dist/build-artifact
-              `
-            });
-            export const project = garn.mkProject({
-              description: "",
-              defaultEnvironment: env,
-            }, {})
-              .addPackage("build", "mv dist/build-artifact $out/build-artifact");
+              const env = garn.mkEnvironment()
+                .addToSandboxSetup(garn.nix.nixStrLit`
```
alexdavid, line 98: I think these methods no longer exist on `Environment`, right?

ts/environment.ts
```diff
@@ -62,11 +65,66 @@ export type Environment = {
...
+      },
+    ],
+  };
+}
+
+/**
+ * Add the common setup of an `Environment` to a given script. This should be
+ * used for `Executable`s and `garn enter`.
+ */
+export const commonScript = (
```
alexdavid, line 106: I'm a little confused by the name here. Shouldn't this be `wrapScriptInEnv` or something?

ts/nix.ts
soenkehahn: Done in a9ba3b57.

CHANGELOG.md
soenkehahn: Changed the changelog entry. Is that better?

alexdavid (approved):

PR #451, opened 2023-11-25
Add CI check that flake files are up to date

soenkehahn:


DavHau (approved):

DavHau:
There still seems to be a bug in this.
If you checkout revision ab0f01fc82c976d4f6c1136d9fb1079b613a77a4 for example, and then run `just update-flakefiles`, you can see that files change, but when it was [tested here](https://github.com/garnix-io/garn/actions/runs/7085431345/job/19281643312), no errors did occur.

DavHau:
> There still seems to be a bug in this. If you checkout revision [ab0f01f](https://github.com/garnix-io/garn/commit/ab0f01fc82c976d4f6c1136d9fb1079b613a77a4) for example, and then run `just update-flakefiles`, you can see that files change, but when it was [tested here](https://github.com/garnix-io/garn/actions/runs/7085431345/job/19281643312), no errors did occur.

Ah nevermind, seems like the problem was that I did not start the fileserver before running `just update-flakefiles`

PR #462, opened 2023-11-28
Add test to verify `docland` for garn works

alexdavid:
`docland` uses an older version of `deno_doc` which doesn't support all the newest features of typescript. This tests verifies that pushing such a change results in a CI error.


ts/internal/docland.test.ts
```diff
@@ -0,0 +1,11 @@
+// This is the dependency file of the repo deployed at doc.deno.land. We import
+// deno_doc's `doc` here to check if our modules are parsable by the verison of
```
soenkehahn, line 2:
```suggestion
// deno_doc's `doc` here to check if our modules are parsable by the version of
```

soenkehahn (approved):

PR #465, opened 2023-11-29
check examples are always up-to-date

DavHau:
- update all flake examples
- check examples are always up-to-date

### DIFF
--- a/justfile
+++ b/justfile
@@ -8,7 +8,7 @@ pre-push: fmt github-ci
 fast-pre-push: fmt update-flakefiles test check-examples typescript-check test-ts
 
 # Run checks that we can’t yet run via the flake
-github-ci: codegen test check-isolated-garn check-examples test-ts
+github-ci: (update-flakefiles "true") codegen test check-isolated-garn check-examples test-ts
 
 fmt: fmt-nix fmt-haskell fmt-typescript hpack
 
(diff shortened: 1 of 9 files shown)


justfile
```diff
@@ -144,16 +144,27 @@ check-examples:
   just run-garn go-http-backend check
   just run-garn monorepo check
 
-update-flakefiles:
+update-flakefiles check="false":
   #!/usr/bin/env bash
+  # set -euo pipefail
```
alexdavid, line 149: Is making this a comment intentional?

justfile
```diff
@@ -8,7 +8,7 @@ pre-push: fmt github-ci
 fast-pre-push: fmt update-flakefiles test check-examples typescript-check test-ts
 
 # Run checks that we can’t yet run via the flake
-github-ci: codegen test check-isolated-garn check-examples test-ts
+github-ci: (update-flakefiles "true") codegen test check-isolated-garn check-examples test-ts
```
alexdavid, line 11: What are your thoughts on having a step that verifies that after running all the tests the repo is still in a clean state? This could simplify `update-flakefiles`. Although I don't feel strongly, TIOLI!

```suggestion
github-ci: update-flakefiles codegen test check-isolated-garn check-examples test-ts check-not-dirty

check-not-dirty:
  git diff-files --quiet
```

alexdavid (approved):
Good catch! It'll be nice to have this caught automatically in the future

justfile
DavHau: No it wasn't. Thanks for the hint. It's fixed now

soenkehahn:
Closing in favor of #451.

PR #466, opened 2023-11-29
add devShell ts test for mkNpmProject

DavHau:


ts/testUtils.ts
```diff
@@ -164,6 +164,37 @@ export const buildPackage = (
   return Deno.readLinkSync(`${dir}/result`);
 };
 
+export const runInDevShell = (
+  devShell: garn.Environment,
+  options: { cmd: string; dir?: string } = { cmd: "true" },
+): void => {
+  const dir = options.dir ?? Deno.makeTempDirSync({ prefix: "garn-test" });
```
alexdavid, line 171: Nit, maybe it's nice to default this the same way we do dir? It's kind of sad that typescript doesn't have great syntax for defaulting object properties

```suggestion
export const runInDevShell = (
  devShell: garn.Environment,
  options: { cmd?: string; dir?: string } = {},
): void => {
  const cmd = options.cmd ?? "true";
  const dir = options.dir ?? Deno.makeTempDirSync({ prefix: "garn-test" });
```

jkarni:
Looks like `check-git-is-unchanged` is failing - maybe you need to run `update-flakefiles`?

PR #474, opened 2023-12-02
Home page Hero

ericyliu:
1. added hero component to home page
2. added `react-syntax-highlighter` package
3. added regular font and made header font weight 300
4. the automated typing code is copied from the original site

The syntax highlighting isn't exactly matching the design, not sure how picky we are about that. Picked a preset that was close enough, but I can try to play around with it to get it closer if that's what we want.

hero-desktop
hero-mobile


alexdavid (approved):
This looks great! I think the syntax highlighting not perfectly matching the designs is fine.

PR #478, opened 2023-12-05
Home page CI section

ericyliu:
Screenshot 2023-12-05 at 1 46 51 PM
Screenshot 2023-12-05 at 1 47 02 PM


website2/src/components/homePage/ci/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+export const backend = garn.haskell
+  .mkHaskellProject({
+    description: "my Haskell backend",
+    executable: "server",
+    compiler: "ghc94",
+    src: ".",
+  })
+  .withDevTools([nixpkgs.hlint])
+  .addCheck("hlint")\`hlint *.hs\`;`;
```
alexdavid, line 18: FYI these examples have changed (see garn.io) However, it probably can be done as a pass at the end before we merge the redesign since they will likely continue changing

```suggestion
const CODE = `import * as garn from "https://garn.io/ts/v0.0.18/mod.ts";
import * as nixpkgs from "https://garn.io/ts/v0.0.18/nixpkgs.ts";

export const backend = garn.haskell
  .mkHaskellProject({
    description: "my Haskell backend",
    executable: "server",
    compiler: "ghc94",
    src: ".",
  })
  .withDevTools([nixpkgs.hlint])
  .addCheck("hlint", "hlint *.hs");`;
```

alexdavid (approved):
:tada:

PR #488, opened 2023-12-12
Documentation page document

ericyliu:
1. still use mdx files for documentation
2. changed out some components in the mdx files for new ones

getting-started
getting-started-mobile


website2/src/app/documentation/[slug]/page.tsx
Comment (alexdavid) [inline website2/src/app/documentation/[slug]/page.tsx:1]: The old website used `docs` for the url. Should we keep the same paths for SEO?

website2/documentation/concepts.mdx
```diff
@@ -0,0 +1,153 @@
+export const info = {
+  name: "concepts",
+  url: "concepts",
```
alexdavid, line 3: I like the fact that we use the file names for the url. I think because we're doing that we can remove these `url` properties, right?

website2/documentation/getting-started.mdx
alexdavid, line 1: Same vein as above, I actually prefer kabob case over snake case, but the current website uses snake case (`/docs/getting_started`). I wonder how much effort is it to add a 301 redirect to the kabob version? If it's too much work though I'd suggest we just keep the snake case.

website2/documentation/concepts.mdx
```diff
@@ -0,0 +1,153 @@
...
+
+- Running your projects main executable during development,
+- Running a code formatter,
+- Running a code generator,
+- Spinning up a database for development,
+- etc.
+
+Here's a small example:
+
+```typescript
```
alexdavid, line 83: There seems to be some `text-shadow` applied to the `<pre>` element on these codeblocks. It makes everything look kind of blurry. Is this applied by `next-mdx-remote`?

alexdavid (approved):

website2/src/app/documentation/[slug]/page.tsx
Comment (ericyliu) [reply website2/src/app/documentation/[slug]/page.tsx:1]: ah thats a fair point i'll change that, i can keep the mdx file the same name too then

website2/documentation/getting-started.mdx
ericyliu: not too much, I can add a redirect

PR #489, opened 2023-12-13
Documentation anchor tags

ericyliu:
adds anchor tags and ids to headers in the documentation

anchor
Screenshot 2023-12-13 at 2 13 59 PM

### DIFF
--- a/website2/src/components/documentationPage/documentHeader/index.tsx
+++ b/website2/src/components/documentationPage/documentHeader/index.tsx
@@ -1,5 +1,5 @@
 import { Text } from "@/components/text";
-import { PropsWithChildren, ReactNode } from "react";
+import { PropsWithChildren, ReactNode, isValidElement } from "react";
 import styles from "./styles.module.css";
 import { Share } from "@/components/icons/share";
 import Link from "next/link";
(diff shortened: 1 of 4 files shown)


website2/src/components/documentationPage/documentHeader/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+      <Link href={`#${id}`} className={styles.link}>
+        <span>{children}</span>
+        <Share className={styles.share} />
+      </Link>
+    </Text>
+  );
+};
+
+const getId = (children: ReactNode): string => {
+  return children?.toString().toLowerCase().split(" ").join("-") || "";
```
alexdavid, line 30: Is it meaningful to call `toString` on a `ReactNode`? I could imagine some cases (like the react node being an array which it is for multiple children) where this might be a bit weird. I might be wrong here, let me know if react uses custom types for this that overload `toString`.

If this doesn't work well, maybe should we restrict `DocumentHeader` to just take `children: string` for now if we expect this to always just be a plain text react node?

Also small nit, by using replace we can target other characters other than spaces which probably doesn't really apply. TIOLI

```suggestion
children.toLowerCase().replace(/[^a-zA-Z0-9]+/g, "-")
```

alexdavid (approved):

website2/src/components/documentationPage/documentHeader/index.tsx
ericyliu: I'll uss replace since that's better. This also applies to Text though, and neither really is locked down to only string. For example, in the mdx file, we have <Garn /> inside a Text

ericyliu: I fixed id generation with a different method, which includes the regex replace. we are now casing on the different types a ReactNode could be

PR #490, opened 2023-12-13
Add Zod typechecking to mdx files

ericyliu:
- now does build time typechecking using zod on react components inside mdx files


website2/src/components/documentationPage/documentHeader/index.tsx
```diff
@@ -1,30 +1,33 @@
...
-  type: "h1" | "h2" | "h3";
-  className?: string;
-};
+const PropSchema = z
+  .object({
+    type: z.enum(["h1", "h2", "h3"]),
+    className: z.string().optional(),
+    children: z.custom(),
+  })
+  .strict();
```
ericyliu, line 15: @alexdavid i would like to call `.strict()` inside `checkPropsRuntime` but I can't get the typing to work out

website2/src/components/documentationPage/documentHeader/index.tsx
```diff
@@ -1,30 +1,31 @@
...
-      className={`${rest.className} ${styles.container}`}
-    >
-      <Link href={`#${id}`} className={styles.link}>
-        <span>{children}</span>
-        <Share className={styles.share} />
-      </Link>
-    </Text>
-  );
-};
+export const DocumentHeader = withPropCheck<z.infer<typeof PropSchema>>(
```
alexdavid, line 15: Can `T` be inferred here from the first argument? If not I think maybe going the other way of making `T` the type of `PropSchema` and using `z.infer` in the signature of the passed in component should let typescript auto-infer this so we don't have to write the generic all the time

alexdavid (approved):
What are your thoughts on putting the calls to `withPropCheck` in the `components` object of `MDXRemote`? We only need them there, right? Or is there an advantage to having the runtime typechecking in the exported components instead?

ericyliu:
> What are your thoughts on putting the calls to `withPropCheck` in the `components` object of `MDXRemote`? We only need them there, right? Or is there an advantage to having the runtime typechecking in the exported components instead?

thats true.. I wanted to colocate the schema with the component itself. but I think saying that it's only going to be a "mdx" concept is fine too

website2/src/components/documentationPage/documentHeader/index.tsx
ericyliu: yea it can be, i'll remove it from places

alexdavid:
Nope don't feel strongly. It's a good point that it might be simpler to keep the schema next to the component

ericyliu:
> Nope don't feel strongly. It's a good point that it might be simpler to keep the schema next to the component

then I'll leave it as is for simplicity's sake

PR #491, opened 2023-12-13
Documentation fix getting started

ericyliu:
redirect only works if we're running the nextjs app as a server and not building a bunch of static pages (which is necessary with github pages). Changing the filename back to `getting_started.mdx` so it matches the current route


ericyliu:
or we could do this https://docs.github.com/en/contributing/writing-for-github-docs/configuring-redirects

alexdavid (approved):
Ahh right that makes sense

PR #495, opened 2023-12-19
Fix SVGs on website

ericyliu:
- changes SVGs to be react components to handle caching better


alexdavid (commented):
One issue with making these react components is they are no longer lazy loaded (and `shape.svg` is quite large). I think we can leave them as SVGs  but just move them out of public, right?

ericyliu:
> One issue with making these react components is they are no longer lazy loaded (and `shape.svg` is quite large). I think we can leave them as SVGs but just move them out of public, right?

So there's 3 ways to load an SVG onto the page:

1. use `next/image`, but then the svg needs to be in the public folder (specific nextjs behavior, trying to optimize image loading). This method also prevents us from interacting with the SVG directly (i.e. with css fill, stroke)

2. import as a react component (but like you said, we lose lazy loading)

3. import with svgr (https://react-svgr.com/docs/next/), but im pretty sure we also lose lazy loading

I think the choices are basically lazy loading (and the img has to be in the public folder) vs being able to interact with the svg. We can make it so shape is an image (since we are not interacting with it), but make the others react components. There is also lazy loading in react if we really want to optimize this, but we might be doing too much work optimizing at this juncture. https://react.dev/reference/react/lazy

alexdavid (approved):

PR #497, opened 2024-01-10
Add analytics to new site

ericyliu:


website2/src/app/layout.tsx
```diff
@@ -12,6 +12,13 @@ export const metadata: Metadata = {
 const RootLayout = ({ children }: { children: React.ReactNode }) => {
   return (
     <html lang="en">
+      <head>
+        <script
+          defer
+          data-domain="garn.io"
```
alexdavid, line 18:
```suggestion
          data-domain="garnix.io"
```

alexdavid (approved):

website2/src/app/layout.tsx
alexdavid: Oops looking at the wrong board, my bad. Disregard

PR #501, opened 2024-01-16
Fix horizontal overflow for SampleCode component

ericyliu:
fixes issue [#687](https://github.com/garnix-io/garnix/issues/687)

- not the most elegant fix, but prevents the horizontal overflow. We could consider making the code in the box smaller incrementally with less horizontal screen space to prevent line wrapping, but that's not a perfect solution either
- also removes some extra padding at the bottom of SampleCode

Screenshot 2024-01-16 at 2 43 21 PM


alexdavid (approved):
Should we add a horizontal scrollbar to long codeblocks? I think I've seen that more often than wrapping, but I don't feel strongly

ericyliu:
that's a good point, it'll look nicer for the parts that are visible, let me change that

PR #503, opened 2024-01-22
Fix website deploy scripts to deploy new version

soenkehahn:


ericyliu (approved):
Seems right,  instead of having website out you're having out from website2

jkarni (approved):

alexdavid (approved):

PR #507, opened 2026-04-01
Fix spelling

jsoref:
This project appears dead, so this PR probably isn't worth making, but it's been on my todo list, so I'm making it as part of my spring cleaning.


jkarni (approved):
This is fantastic! Thanks so much. 

Re: the project being dead. We've actually started a [rewrite](https://github.com/garnix-io/garn2) that we work on in bursts of activity, but it'll likely inherit the majority of documentation and of the website, so this is great.

